PR #1, opened 2018-08-02
Added support to upgrade/downgrade subscription

AqvinSaldanha:
Renamed "reason" to message in error JSON object to make it compatible with the Error class defined in Purchases.cs (Unity plugin)
purchasesunity.aar built with above changes


jeiting:
Thanks Aqvin!

PR #9, opened 2019-03-29
More attribution networks and a way to set finishtransactions in ios

vegaro:


Subtester/Assets/PurchasesListener.cs
```diff
@@ -20,6 +20,9 @@ void Start()
         CreateButton("Switch Username", () => SwitchUser(), 200);
         
         CreateButton("Send Attribution", () => SendAttribution(), 300);
+
+        Purchases purchases = GetComponent<Purchases>();
+        purchases.FinishTransactions(false);
```
jeiting: shouldn't be here

PR #19, opened 2020-04-18
[FINAL] Moves inner classes to their own files

vegaro:
`Purchases.cs` was getting out of hand so I moved all the inner classes to their own files


RevenueCat/Scripts/Offering.cs
```diff
@@ -0,0 +1,58 @@
...
+        public readonly string Identifier;
+        public readonly string ServerDescription;
+        public readonly List<Package> AvailablePackages;
+        [CanBeNull] public readonly Package Lifetime;
+        [CanBeNull] public readonly Package Annual;
+        [CanBeNull] public readonly Package SixMonth;
+        [CanBeNull] public readonly Package ThreeMonth;
+        [CanBeNull] public readonly Package TwoMonth;
+        [CanBeNull] public readonly Package Monthly;
+        [CanBeNull] public readonly Package Weekly;
```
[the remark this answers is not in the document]
aboedo: or is the criteria: 
- `[CanBeNull]` + `Package?` if it can validly be _assigned_ to null
- `[CanBeNull]` + `Package` if it can only be null if the value has never been set

RevenueCat/Scripts/Product.cs
```diff
@@ -0,0 +1,36 @@
...
+public partial class Purchases
+{
+    public class Product
+    {
+        public string title;
+        public string identifier;
+        public string description;
+        public float price;
+        public string priceString;
+        [CanBeNull] public string currencyCode;
```
aboedo, line 12: same question regarding nullability and `?`

RevenueCat/Scripts/Purchases.cs
```diff
@@ -884,181 +404,5 @@ private static void ReceivePurchaserInfoMethod(string arguments, PurchaserInfoFu
...
-        public float price;
-        public string price_string;
-        [CanBeNull] public string currency_code;
-        public float intro_price;
-        public string intro_price_string;
-        public string intro_price_period;
-        public string intro_price_period_unit;
-        [CanBeNull] public int intro_price_period_number_of_units;
-        [CanBeNull] public int intro_price_cycles;
-    }
```
aboedo, line 1063: 🎉 👏 🎉 👏 🎉 👏 🎉 👏

vegaro:
Regarding the `[canbenull]` and `?`. If I add the ? to string, Rider removes it and adds the [CanBeNull] automatically 🤔 

I think it makes sense then to use [CanBeNull] only.

vegaro:
Ok, I read more about `[CanBeNull]` and `?`. `DateTime?` is syntactic sugar to doing `Nullable<DateTime>`. It wraps a datetime in a Nullable, which forces the developer to get the value and to check for the value to be not null. [CanBeNull] is an annotation similar to `nullable` in xcode.

Things that are inherently nullable (classes) like string, can't be wrapped in a nullable (you can't do `string?`). We can do `DateTime?` because it's a struct. Doing `[CanBeNull]` and also `DateTime?` means that the Nullable wrapper that wraps the DateTime can also be null, which is not what we want.

PR #24, opened 2020-05-29
maven dependency and new headers

aboedo:
Replaces duplicated files for purchases-hybrid-common with a maven dependency, adds new headers for `platformFlavor` and `platformFlavorVersion`. 

### Requirements: 
- [ ] testing on device (having a few issues with Android Unity testing right now)


update-sdks.sh
```diff
@@ -1,23 +0,0 @@
-#!/bin/bash
```
vegaro, line 1: We can clean up the RELEASING.md a little bit too 🎊

aboedo: good catch. I have an android device set to arrive today, when it gets here I'll do a pass of testing and I'll clean up RELEASING.md

aboedo: coming back to this, what _should_ be updated in RELEASING.md?

aboedo: @vegaro I'm removing the hold and merging for now and I can follow up with any needed changes to RELEASING.md in a separate PR

PR #25, opened 2020-06-26
Feature/bump native sdks

aboedo:
Updates Unity SDK for the following: 
- `setProxyURLString`
- `userDefaultsSuiteName` in setup
- `originalPurchaseDate` in purchaser info
- `managementURL` in purchaserInfo

Also: 
- Updates play services resolver to the latest version
- Moves `PurchasesWrapper.java` and `AdvertisingIdClient.java` to `Revenuecat/Plugins/Android/`, so that we no longer need to call `build_and_install` any time those are modified 🎉  (Thanks for the assist @vegaro !)
- Deletes `Android` folder since it's no longer needed
- Fixes a bug in present in `develop` where we were missing an import for `PlatformInfo`


RevenueCat/Plugins/Android/PurchasesWrapper.java
```diff
@@ -0,0 +1,331 @@
+package com.revenuecat.purchasesunity;
+
+import android.util.Log;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.revenuecat.purchases.PurchaserInfo;
+import com.revenuecat.purchases.Purchases;
+import com.revenuecat.purchases.PlatformInfo;
```
aboedo, line 10: this line was missing from https://github.com/RevenueCat/purchases-unity/pull/24/files

RevenueCat/Plugins/Android/PurchasesWrapper.java
```diff
@@ -0,0 +1,331 @@
+package com.revenuecat.purchasesunity;
```
aboedo, line 1: this file isn't new, it was moved

RevenueCat/Plugins/Android/PurchasesWrapper.java
```diff
@@ -0,0 +1,331 @@
...
+        public void onReceived(@NonNull PurchaserInfo purchaserInfo) {
+            sendPurchaserInfo(MappersKt.map(purchaserInfo), RECEIVE_PURCHASER_INFO);
+        }
+    };
+
+    public static void setup(String apiKey,
+                             String appUserId,
+                             String gameObject_,
+                             boolean observerMode,
+                             String userDefaultsSuiteName) {
```
aboedo, line 56: this last param is new, it's just ignored. I debated whether to even pass it all the way here, but I figured it was better for the actual java layer to determine what it needs or what it doesn't, since the intermediate layers should be as dumb as possible.

RevenueCat/Plugins/Android/PurchasesWrapper.java
```diff
@@ -0,0 +1,331 @@
...
+        });
+    }
+
+    public static void setDebugLogsEnabled(boolean enabled) {
+        CommonKt.setDebugLogsEnabled(enabled);
+    }
+
+    public static void setProxyURL(String proxyURL) {
+        CommonKt.setProxyURLString(proxyURL);
+    }
```
aboedo, line 208: this is the other new thing on this file

RevenueCat/Scripts/PurchaserInfo.cs
```diff
@@ -94,6 +99,8 @@ public override string ToString()
                    $"{nameof(RequestDate)}: {RequestDate}, " +
                    $"{nameof(AllExpirationDates)}: {AllExpirationDates}, " +
                    $"{nameof(AllPurchaseDates)}: {AllPurchaseDates}, " +
+                   $"{nameof(OriginalPurchaseDate)}: {OriginalPurchaseDate}, " +
```
aboedo, line 102: coming to android soon for compatibility with users who purchased on iOS

RevenueCat/Scripts/Purchases.cs
```diff
@@ -164,6 +167,14 @@ public void SetDebugLogsEnabled(bool logsEnabled)
         _wrapper.SetDebugLogsEnabled(logsEnabled);
     }
 
+    // ReSharper disable once UnusedMember.Global
+    public void SetProxyURL(string proxyURL)
+    {
+        _wrapper.SetProxyURL(proxyURL);
+        Setup(string.IsNullOrEmpty(appUserID) ? null : appUserID);
+        GetProducts(productIdentifiers, null);
```
aboedo, line 175: this bit is up for discussion. 
On other SDKs, we would ask devs to call `SetProxyURL` before setup. but here, setup isn't dev-controlled, it happens automatically from the params passed into the GUI. `GetProducts` is also called automatically. 
Since if you're using a proxy, you'd miss those calls, I'm making them here. 

An alternative approach would have been for this to be an extra param in the GUI, like `userDefaultsSuiteName`. However, I think we didn't want to expose that value too much, since it's really only meant for very specific usage. 

Come to think about it, this does mean, though, that your app may actually perform a request to the revenueCat servers before hitting the proxy. 
So if we want to be absolutely sure that our servers are never hit directly, it'll have to be a param in the GUI. 

Thoughts?

Android/build_and_install.sh
```diff
@@ -1,4 +0,0 @@
-#!/bin/bash
```
vegaro, line 1: Let's update the RELEASING.md too to indicate that we don't need to run this anymore

PurchasesWrapper.java
```diff
@@ -48,7 +48,7 @@ public void onReceived(@NonNull PurchaserInfo purchaserInfo) {
         }
     };
 
-    public static void setup(String apiKey, String appUserId, String gameObject_, boolean observerMode) {
+    public static void setup(String apiKey, String appUserId, String gameObject_, boolean observerMode, string userDefaultsSuiteName) {
         gameObject = gameObject_;
         PlatformInfo platformInfo = new PlatformInfo(PLATFORM_NAME, PLUGIN_VERSION);
         CommonKt.configure(UnityPlayer.currentActivity, apiKey, appUserID, observerMode, platformInfo);
```
vegaro, line 54: There's a compilation error here, since it can't find `appUserID`. The parameter name is `appUserId`. Why haven't we noticed this before?

vegaro: Oh actually, this file is in the root of the repository. We should remove it. I see the `appUserId` fix is done in the other `PurchasesWrapper.java`

RevenueCat/Plugins/Editor/RevenueCatDependencies.xml
```diff
@@ -2,10 +2,10 @@
 <dependencies>
     <androidPackages>
         <androidPackage spec="org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.2.61" />
-        <androidPackage spec="com.revenuecat.purchases:purchases:3.1.0" />
-        <androidPackage spec="com.revenuecat.purchases:purchases-hybrid-common:1.1.2" />
+        <androidPackage spec="com.revenuecat.purchases:purchases:3.2.0" />
```
vegaro, line 5: I think this one can be removed. The latest `purchases-hybrid-common` should include it. Let's test it before we release but I removed it from Flutter. We can get weird issues if the version that `purchases-hybrid-common` uses differs from the one we declare here.

RevenueCat/Scripts/PurchaserInfo.cs
```diff
@@ -24,9 +24,12 @@ public class PurchaserInfo
         public DateTime FirstSeen;
         public string OriginalAppUserId;
         public DateTime RequestDate;
+        [CanBeNull] public DateTime OriginalPurchaseDate;
```
vegaro, line 27: I believe this should be `DateTime?` instead of `CanBeNull` because `CanBeNull` indicates that the pointer can be null and `DateTime?` wraps it in an `Optional`, like `LatestExpirationDate`. I always get confused with this though

RevenueCat/Scripts/PurchaserInfo.cs
```diff
@@ -52,6 +55,8 @@ public PurchaserInfo(JSONNode response)
             FirstSeen = FromUnixTime(response["firstSeenMillis"].AsLong);
             OriginalAppUserId = response["originalAppUserId"];
             RequestDate = FromUnixTime(response["requestDateMillis"].AsLong);
+            OriginalPurchaseDate = FromUnixTime(response["originalPurchaseDateMillis"].AsLong);
```
vegaro, line 58: Since `originalPurchaseDateMillis` can be null, you should do something similar to `LatestExpirationDate`. Otherwise it will initialize an empty `DateTime` if `response["originalPurchaseDateMillis"]` is null. At least that's how it worked before I changed the JSON parser. You should double check.

RevenueCat/Scripts/Purchases.cs
vegaro: Is there a way in Unity to add like a dropdown to the GUI? So that we can put this under an `Advanced` option under the `userDefaultsSuiteName` in the GUI? I have no clue. I will check.

vegaro: Ok, I checked and there's no way to do like a dropdown type of thing that hides some Advanced attributes, but we could add a header that says Advanced
https://unity3d.college/2017/05/22/unity-attributes/

Subtester/Assets/PlayServicesResolver/Editor/play-services-resolver_v1.2.135.0.txt
```diff
@@ -1,8 +0,0 @@
-Assets/PlayServicesResolver/Editor/Google.IOSResolver_v1.2.135.0.dll
-Assets/PlayServicesResolver/Editor/Google.IOSResolver_v1.2.135.0.dll.mdb
-Assets/PlayServicesResolver/Editor/Google.JarResolver_v1.2.135.0.dll
-Assets/PlayServicesResolver/Editor/Google.JarResolver_v1.2.135.0.dll.mdb
-Assets/PlayServicesResolver/Editor/Google.VersionHandler.dll
-Assets/PlayServicesResolver/Editor/Google.VersionHandler.dll.mdb
-Assets/PlayServicesResolver/Editor/Google.VersionHandlerImpl_v1.2.135.0.dll
-Assets/PlayServicesResolver/Editor/Google.VersionHandlerImpl_v1.2.135.0.dll.mdb
```
vegaro, line 8: This got removed because you upgraded the plugin, but where are the references to the newer version? 🤔

RevenueCat/Scripts/Purchases.cs
aboedo: great idea! I'm adding it, will test and update the PR
image

aboedo: tested and working 🎉

aboedo:
@vegaro changes implemented, ready for another pass

Subtester/Assets/PlayServicesResolver/Editor/play-services-resolver_v1.2.135.0.txt
vegaro: oh wow it means we don't have to download it manually anymore!!!!! Does it mean we don't have to do this either? https://github.com/googlesamples/unity-jar-resolver#plugin-redistribution

Like we don't have to include the unity-jar-resolver package when exporting our package anymore?

PR #27, opened 2020-07-30
Prepares version 2.2.0

vegaro:
I added a script that creates the unitypackages. I had to remove the Unity Package Manager dependency because we can't export the ExternalDependencyManager within the unitypackage, so I brought it back to the plugins folder. I opened https://github.com/googlesamples/unity-jar-resolver/issues/393 for more clarification on this.

We never released `2.1.0`, the release was ready but we started accumulating changes and we never got to release it. 

I added the `2.1.0` tag and will make this one `2.2.0`.

`purchasesunity.aar` and `purchasesunity-sources.jar` are leftovers from when we removed the Android project, we don't need it anymore.

The contents of the `ObserverMode` folder are used when creating the special unitypackage for observer mode.

I renamed the `billingclient.billing-2.0.3-no-service.aar` to `billingclient.billing-no-service` so that when we update it, it gets replaced in people's projects.


CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
...
+  - Android downloads the common code from Maven
+  - Removed a bunch of scripts
+  - Removed the Android project and added the files to the plugin folder
+  - Added platform version and name to the http headers
+  - Added a script to create the unitypackage
+  - Moved inner classes to their own files
+  - Use SimpleJSON instead of JSONUtility
+- Adds proxyURL property, useful for kids category apps, so that they can set up a proxy to send requests through. Do not use this unless you've talked to RevenueCat support about it.
+- Adds userDefaultsSuiteName. Set this to use a specific NSUserDefaults suite for RevenueCat. This might be handy if you are deleting all NSUserDefaults in your app and leaving RevenueCat in a bad state.
+- Update to ExternalDependencyManager 1.2.156
```
aboedo, line 14: let's maybe move these up to the top, since they're the most likely to be interesting for devs

scripts/create-unity-package.sh
```diff
@@ -0,0 +1,92 @@
...
+if [[ $? -ne 0 ]]; then
+  echo "😱 Exporting has failed! Checkout exportlogs.txt"
+  echo "😱 Unity shouldn't be running when executing the script."
+  exit $return_code
+else
+  echo "🚀 Exporting finished for the regular package! Path: $PACKAGE" 
+fi
+
+echo "📦 Moving com.android.billingclient.billing-no-service.aar to Revenuecat/Plugins/Android"
+cp $PWD/ObserverMode/com.android.billingclient.billing-no-service.aar $PWD/RevenueCat/Plugins/Android
```
aboedo, line 66: `$PWD/RevenueCat/Plugins` is repeated quite a few times, maybe it's worth moving it to a variable

PR #28, opened 2020-08-05
Preparing release 2.2.1

vegaro:
When I prepared the unitypackage, the iOS folder still had the .h and .m files from before we migrated to use cocoapods, so trying to compile iOS would throw a duplicated files error.


vegaro:
I merged it because there's basically no changes and it's blocking since the current release doesn't work atm

PR #32, opened 2020-10-23
upgraded unity version in subtester app

aboedo:
upgraded unity version in subtester app. 
With these changes, we no longer have to manually set the swift language version or 'enable modules' every time we export the project.


vegaro (approved):
Awesome, I will update my Unity

PR #38, opened 2020-12-29
Release/3.0.0

aboedo:
## 3.0.0
- removes deprecated `MakePurchase`, replaced by `PurchaseProduct`
  - iOS: 
      - added new method, `syncPurchases`, that enables syncing the purchases in the local receipt with the backend without risking a password prompt. The method was already available on Android.
      - added a new method, `presentCodeRedemptionSheet`, for offer codes redemption.
      - fixed a bug where values for dates in milliseconds were actually in seconds (https://github.com/RevenueCat/purchases-hybrid-common/pull/62)
 - Bumped `purchases-hybrid-common` to 1.5.0 [Changelog here](https://github.com/RevenueCat/purchases-hybrid-common/releases/1.5.0)
 - Bumped `purchases-ios` to 3.9.2 [Changelog here](https://github.com/RevenueCat/purchases-ios/releases/3.9.2)
 - Bumped `purchases-android` to 4.0.1 [Changelog here](https://github.com/RevenueCat/purchases-android/releases/4.0.1)


CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 3.0.0
+- removes deprecated `MakePurchase`, replaced by `PurchaseProduct`
+  - iOS:
```
vegaro, line 3: I think this indentation is wrong

Subtester/ProjectSettings/AndroidResolverDependencies.xml
```diff
@@ -1,6 +1,6 @@
 <dependencies>
   <packages>
-    <package>com.revenuecat.purchases:purchases-hybrid-common:1.4.4</package>
```
aboedo, line 3: not sure why this wasn't `1.4.5` before, we might have missed that when doing the previous update

vegaro: I think it's because the project was never ran and it never got updated. This is like a `package.json.lock`, so if the project is never ran with the new dependencies, it won't get updated. That's my guess. Thanks for updating it!

vegaro (approved):

PR #41, opened 2021-03-10
Amazon support

vegaro:


aboedo:
@vegaro we should be able to close this one out in favor of 4.0.0, right?

PR #48, opened 2021-04-16
Feature: identity v3

aboedo:
Adds new methods: 
- `logIn`
- `logOut`

Deprecates: 
- `allowSharingStoreAccount`
- `identify`
- `createAlias`
- `reset`

### DIFF
--- a/RevenueCat/Plugins/Android/PurchasesWrapper.java
+++ b/RevenueCat/Plugins/Android/PurchasesWrapper.java
@@ -34,6 +34,8 @@ public class PurchasesWrapper {
     private static final String CREATE_ALIAS = "_createAlias";
     private static final String RECEIVE_PURCHASER_INFO = "_receivePurchaserInfo";
     private static final String RESTORE_TRANSACTIONS = "_restoreTransactions";
+    private static final String LOG_IN = "_logIn";
+    private static final String LOG_OUT = "_logOut";
     private static final String IDENTIFY = "_identify";
     private static final String RESET = "_reset";
     private static final String GET_OFFERINGS = "_getOfferings";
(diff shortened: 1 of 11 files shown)


aboedo:
I wasn't able to test this one on device in Android, unfortunately. I did test iOS on device, though. 
I ran into some complications when trying to run from a git branch or even from local copies of `purchases-android` and `purchases-hybrid-common`. I think the issues were related to Unity's unique settings for Android, where it uses a lot of old stuff and also sets up the android purchases libraries as `.aar`s. 

I'm open to suggestions, but in the meantime I'm marking this as HOLD, and I think we can just wait to merge it until there's a released version of `purchases-android` and `purchases-hybrid-common` with identity v3

aboedo:
should be fine to review, though

vegaro (approved):

aboedo:
@vegaro we now have to update the billing client dependency to 4.0.0, correct?

aboedo:
answering myself here - we're setting the billing client _through_ our dependency to `purchases-hybrid-common` (and in turn, `purchases-android`), so we don't need to do anything here, other than use the external dependency manager to resolve the versions.

aboedo:
I finally managed to test this on Android, now that we have a released version of `purchases-hybrid-common` out, and fixed a couple of bugs, so I'll ask for a re-revewi

RevenueCat/Plugins/Android/PurchasesWrapper.java
```diff
@@ -395,6 +406,29 @@ private static void sendErrorPurchase(ErrorContainer errorContainer) {
...
+    @NonNull
+    private static OnResult getLogInListener(final String method) {
+        return new OnResult() {
+            @Override
+            public void onReceived(Map<String, ?> map) {
+                JSONObject jsonObject = new JSONObject();
+                try {
+                    Map<String, ?> purchaserInfoMap = (Map<String, ?>)map.get("purchaserInfo");
+                    jsonObject.put("purchaserInfo", MappersHelpersKt.convertToJson(purchaserInfoMap));
+                    jsonObject.put("created", (Boolean)map.get("created"));
```
aboedo, line 418: since we control the data types, the casting here should be safe. 

However, let me know if there's a better way to do this

beylmk: i think it should be fine, if we wanted to be really safe we could catch an InvalidCastException 🤷‍♀️

aboedo: good call, added

PR #50, opened 2021-06-03
Adds script to enable local build

vegaro:
I was working with @beylmk trying to figure out how to build pointing to a local `purchases-hybrid-common` and realized how painful it is. 

I added a script that can be run as:

```
❯ ./prepareLocalDevelopment.sh ~/Documents/exported_project $HOME/Development/repos/purchases-hybrid-common/android                                                     
```

Being the first argument the Unity exported project and the second one the local folder containing the android project of `purchases-hybrid-common`.

The script will:

- Point to the local `purchases-hybrid-common`
- Match the android tools version between both projects
- Enable androidx and jetifier.

Apart from that, when trying to compile with my local version of purchases-hybrid-common, I noticed androidx annotations are not being added as a dependency automatically, which makes sense to me since they are added as `implementation` to `purchases-hybrid-common`, so I added an explicit dependency to `androidx.annotations`.


Subtester/Assets/Plugins/Android/mainTemplate.gradle
```diff
@@ -0,0 +1,73 @@
...
+}
+// Android Resolver Repos End
+apply plugin: 'com.android.library'
+**APPLY_PLUGINS**
+
+dependencies {
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
+// Android Resolver Dependencies Start
+    implementation 'androidx.annotation:annotation:[1.2.0]' // Assets/RevenueCat/Plugins/Editor/RevenueCatDependencies.xml:5
+    implementation 'com.revenuecat.purchases:purchases-hybrid-common:[1.6.2]' // Assets/RevenueCat/Plugins/Editor/RevenueCatDependencies.xml:4
```
vegaro, line 29: Previously, ED4MU would download all aars from the all transitive dependencies and include each of them as `implementation`. This way (with the mainTemplate.gradle) ED4MU adds the dependencies here now instead of downloading all the transitive aars. 

This is more intuitive and easier to work with than having to deal with all the aars. Although, I think we could use includeBuild with both ways, I think I prefer this more.

This is only part of our Subtester project and not part of the plugin that's released.

prepareLocalDevelopment.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+PROJECT_DIRECTORY=$1
+LOCAL_PURCHASES_HYBRID_COMMON_DIRECTORY=$2
+SETTINGS_GRADLE_FILE=$PROJECT_DIRECTORY/settings.gradle
```
vegaro, line 5: I used positional arguments for now, but happy to change.

prepareLocalDevelopment.sh
```diff
@@ -0,0 +1,15 @@
...
+
+echo "" >> $PROJECT_DIRECTORY/settings.gradle
```
vegaro, line 7: This is so the next line is guaranteed to be in a new line. Otherwise, it would append to the contents of the latest line instead, and since there's no empty line at the end of the file, it just appends it like this:

```
include ':launcher', ':unityLibrary'includeBuild('/Users/cesardelavega/Documents/Development/repos/purchases-hybrid-common/android')
```

beylmk (approved):
🥳 thank you!

prepareLocalDevelopment.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
beylmk, line 1: maybe we should name this prepareAndroidLocalDevelopment since it doesn't do ios?

beylmk:
:( it didn't work for me, trying to figure out why and will let you know.

prepareLocalDevelopment.sh
```diff
@@ -0,0 +1,15 @@
...
+
+echo "" >> $PROJECT_DIRECTORY/settings.gradle
+echo "includeBuild('$LOCAL_PURCHASES_HYBRID_COMMON_DIRECTORY')" >> $PROJECT_DIRECTORY/settings.gradle
+
+android_tools_version=$(cat $LOCAL_PURCHASES_HYBRID_COMMON_DIRECTORY/build.gradle | grep ext.android_tools | awk '{print($3)}' | sed "s/'//g")
+sed -i .bck s/3.4.0/$android_tools_version/ $PROJECT_DIRECTORY/build.gradle
+
+echo "" >> $PROJECT_DIRECTORY/gradle.properties
+echo "android.useAndroidX=true" >> $PROJECT_DIRECTORY/gradle.properties
+echo "android.enableJetifier=true" >> $PROJECT_DIRECTORY/gradle.properties
```
beylmk, line 15: could we run a gradle sync after?

Subtester/Assets/Plugins/Android/mainTemplate.gradle
```diff
@@ -0,0 +1,73 @@
+// GENERATED BY UNITY. REMOVE THIS COMMENT TO PREVENT OVERWRITING WHEN EXPORTING AGAIN
```
aboedo, line 1: could you elaborate on this? do we have to remember to remove the comment before exporting every time?

prepareLocalDevelopment.sh
aboedo: very minor, but I think bash scripts are more often than not named using snake case

prepareLocalDevelopment.sh
aboedo: perhaps include that as a comment?

prepareLocalDevelopment.sh
aboedo: also, let's put it in a folder to keep the root folder clean, especially since this is only for development purposes and not for the plugin itself

aboedo (commented):
this is a noble quest indeed.

Subtester/Assets/Plugins/Android/mainTemplate.gradle
vegaro: This comment is automatic too, I didn't write it. We don't have to do anything, ED4MU should update the versions here automatically, but if it doesn't doing a Resolve should.

prepareLocalDevelopment.sh
vegaro: There's no gradle wrapper in the exported project so unless there's a gradle installed in the machine the script can't do the gradle build 🤷‍♂️

aboedo (approved):

PR #51, opened 2021-06-10
Adds required files to create the npm package

vegaro:
Leaving as draft until we decide if we reorganize the repo

Can be installed by adding this as a dependency:
`https://github.com/RevenueCat/purchases-unity.git#package-registry`


vegaro:
This is blocked by reorganizing the plugin so I will close until that gets done

PR #59, opened 2021-08-18
iOS Promotional Offers support 

vegaro:
Adds support for iOS Promo offers

- [x] Test iOS
- [x] Test Android


RevenueCat/Scripts/Product.cs
```diff
@@ -18,7 +19,12 @@ public class Product
         public string introPricePeriodUnit;
         public int introPricePeriodNumberOfUnits;
         public int introPriceCycles;
-
+        /// <summary>
+        /// Collection of iOS promotional offers for a product. Null for Android.
+        /// </summary>
+        /// <returns></returns>
+        [CanBeNull] public Discount[] Discounts;
```
vegaro, line 26: Made this uppercase D although the rest of the properties are lowercase. They should all be uppercase since that's the convention. We should do the change in the near future.

aboedo: that'll require a major release, right?

aboedo: if so, I'm not sure it's best to have one with a proper format and the rest with lowercase, it might just be better to have them all consistently wrong and then doing a single pass to get them all right

aboedo (approved):

RevenueCat/Scripts/Product.cs
vegaro: Ok, I wasn't very comfortable with making just this one uppercase and I was hoping someone had an opinion on it. I will make it lowercase :)

PR #73, opened 2021-12-02
Amazon Unity IAP Support (Based on 3.2.0)

vegaro:
Since Unity IAP hasn't released a version with BillingClient 4 or the new Amazon AppStore SDK, devs that want to use our SDK in observer mode with Unity IAP need to use version 3.2.0 of our SDK. This branch adds the Amazon changes to that version.


aboedo:
@vegaro we should be able to close this one out in favor of 4.0.0, right?

vegaro:
Yeah, I think we are fine now that we have the `unityiap` dependency in place. Although this branch adds this function `SyncObserverModeAmazonPurchase ` https://github.com/RevenueCat/purchases-unity/pull/73/files#diff-08a4dbb2ecb8512aa7a4e475bd3f4269e4943e99e39262b4c13944af0db6142eR257 which is not in main. But it can be closed since we can still check the implementation when we add the function

aboedo:
great catch, I'll add `SyncObserverModeAmazonPurchase`

aboedo:
Looking at this again, the [method is in main here](https://github.com/RevenueCat/purchases-unity/blob/main/RevenueCat/Scripts/Purchases.cs#L297), is there anything else I'm missing? 
I noticed that it's not in API tester, I'm adding it to it now

vegaro:
My bad then, for some reason I looked for it and didn't see it 🤷‍♂️  I am closing this

PR #78, opened 2021-12-29
add new fields for separate API keys

aboedo:
Opening up as a draft to get feedback. 

This adds separate fields for API keys. 
It doesn't remove the old field, but instead marks it as deprecated. 

Notes: 
- the deprecated field will show up as a build warning, but no warnings show up in the editor. So I added the word `deprecated` to its name. 
- I'm automatically switching which field to use based on the current platform. We don't do this for other platforms, but usually you'd get a chance to select which API key you're using in code, and in Unity this is done via editor.
- I renamed the old field but also added a `FormerlySerializedAs` attribute, which will ensure existing projects won't break, since references to the old field will continue to work. 

Screen Shot 2021-12-29 at 5 54 20 PM

[sc-12104]

### DIFF
--- a/RevenueCat/Scripts/Purchases.cs
+++ b/RevenueCat/Scripts/Purchases.cs
@@ -1,4 +1,5 @@
 using UnityEngine;
+using UnityEngine.Serialization;
 using System;
 using System.Collections.Generic;
 using RevenueCat.SimpleJSON;
@@ -41,9 +42,21 @@ public partial class Purchases : MonoBehaviour
     /// </summary>
     public delegate void GetPaymentDiscountFunc(PaymentDiscount paymentDiscount, Error error);
 
-    [Tooltip("Your RevenueCat API Key. Get from https://app.revenuecat.com/")]
+    [FormerlySerializedAs("revenueCatAPIKey")]
+    [ObsoleteAttribute("This property is obsolete. Use revenueCatAPIKeyIOS and revenueCatAPIKeyAndroid instead.", false)]
+    [Tooltip("(DEPRECATED) RevenueCat API Key. Get from https://app.revenuecat.com/. " +
+             "This property is obsolete. " +
+             "Use revenueCatAPIKeyIOS and revenueCatAPIKeyAndroid instead.")]
     // ReSharper disable once InconsistentNaming
-    public string revenueCatAPIKey;
+    public string deprecatedLegacyRevenueCatAPIKey;
+
+    [Tooltip("RevenueCat API Key specifically for iOS. Get from https://app.revenuecat.com/")]
+    // ReSharper disable once InconsistentNaming
+    public string revenueCatAPIKeyIOS;
+
+    [Tooltip("RevenueCat API Key specifically for Android. Get from https://app.revenuecat.com/")]
+    // ReSharper disable once InconsistentNaming
+    public string revenueCatAPIKeyAndroid;
 
     [Tooltip(
         "App user id. Pass in your own ID if your app has accounts. If blank, RevenueCat will generate a user ID for you.")]
@@ -88,7 +101,16 @@ public partial class Purchases : MonoBehaviour
     // Call this if you want to reset with a new user id
     private void Setup(string newUserId)
     {
-        _wrapper.Setup(gameObject.name, revenueCatAPIKey, newUserId, observerMode, userDefaultsSuiteName);
+        var apiKey = deprecatedLegacyRevenueCatAPIKey;
+        if (String.IsNullOrEmpty(apiKey))
+        {
+            if (Application.platform == RuntimePlatform.IPhonePlayer)
+                apiKey = revenueCatAPIKeyIOS;
+            else if (Application.platform == RuntimePlatform.Android)
+                apiKey = revenueCatAPIKeyAndroid;
+        }
+        
+        _wrapper.Setup(gameObject.name, apiKey, newUserId, observerMode, userDefaultsSuiteName);
     }
 
     private GetProductsFunc ProductsCallback { get; set; }
--- a/RevenueCat/Scripts/Purchases.cs
+++ b/RevenueCat/Scripts/Purchases.cs
@@ -43,10 +43,10 @@ public partial class Purchases : MonoBehaviour
     public delegate void GetPaymentDiscountFunc(PaymentDiscount paymentDiscount, Error error);
 
     [FormerlySerializedAs("revenueCatAPIKey")]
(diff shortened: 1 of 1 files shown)


RevenueCat/Scripts/Purchases.cs
```diff
@@ -88,7 +101,16 @@ private void Start()
     // Call this if you want to reset with a new user id
     private void Setup(string newUserId)
     {
-        _wrapper.Setup(gameObject.name, revenueCatAPIKey, newUserId, observerMode, userDefaultsSuiteName);
+        var apiKey = deprecatedLegacyRevenueCatAPIKey;
+        if (String.IsNullOrEmpty(apiKey))
+        {
+            if (Application.platform == RuntimePlatform.IPhonePlayer)
```
NachoSoto, line 107: I wonder if this should take precedence over `deprecatedLegacyRevenueCatAPIKey`? So only check that iff `revenueCatAPIKeyIOS` or `revenueCatAPIKeyAndroid` aren't defined?

NachoSoto (commented):
Nice!!

RevenueCat/Scripts/Purchases.cs
```diff
@@ -41,9 +42,21 @@ public partial class Purchases : MonoBehaviour
...
+    [Tooltip("(DEPRECATED) RevenueCat API Key. Get from https://app.revenuecat.com/. " +
+             "This property is obsolete. " +
+             "Use revenueCatAPIKeyIOS and revenueCatAPIKeyAndroid instead.")]
     // ReSharper disable once InconsistentNaming
-    public string revenueCatAPIKey;
+    public string deprecatedLegacyRevenueCatAPIKey;
+
+    [Tooltip("RevenueCat API Key specifically for iOS. Get from https://app.revenuecat.com/")]
+    // ReSharper disable once InconsistentNaming
+    public string revenueCatAPIKeyIOS;
```
beylmk, line 55: maybe we should name these `revenueCatAPIKeyApple`/`Google`, since the apple one doesn't have to be ios?

RevenueCat/Scripts/Purchases.cs
```diff
@@ -41,9 +42,21 @@ public partial class Purchases : MonoBehaviour
     /// </summary>
     public delegate void GetPaymentDiscountFunc(PaymentDiscount paymentDiscount, Error error);
 
-    [Tooltip("Your RevenueCat API Key. Get from https://app.revenuecat.com/")]
+    [FormerlySerializedAs("revenueCatAPIKey")]
+    [ObsoleteAttribute("This property is obsolete. Use revenueCatAPIKeyIOS and revenueCatAPIKeyAndroid instead.", false)]
+    [Tooltip("(DEPRECATED) RevenueCat API Key. Get from https://app.revenuecat.com/. " +
```
beylmk, line 47: might be good to add that legacy keys will continue working

RevenueCat/Scripts/Purchases.cs
beylmk, line 107: does unity not support other apple platforms?

RevenueCat/Scripts/Purchases.cs
```diff
@@ -88,7 +101,16 @@ private void Start()
     // Call this if you want to reset with a new user id
```
beylmk, line 101: is that comment still valid? wouldn't that be login now? i see this is private anyway but might be confusing

RevenueCat/Scripts/Purchases.cs
vegaro: We don't support MacOS at the moment. But I think the app could be built for MacOS using Catalyst and use the same API Key right?

RevenueCat/Scripts/Purchases.cs
```diff
@@ -88,7 +101,16 @@ private void Start()
...
     private void Setup(string newUserId)
     {
-        _wrapper.Setup(gameObject.name, revenueCatAPIKey, newUserId, observerMode, userDefaultsSuiteName);
+        var apiKey = deprecatedLegacyRevenueCatAPIKey;
+        if (String.IsNullOrEmpty(apiKey))
+        {
+            if (Application.platform == RuntimePlatform.IPhonePlayer)
+                apiKey = revenueCatAPIKeyIOS;
+            else if (Application.platform == RuntimePlatform.Android)
+                apiKey = revenueCatAPIKeyAndroid;
```
vegaro, line 110: When adding Amazon we can check if `useAmazon` is true here and use its API Key.

vegaro (approved):

RevenueCat/Scripts/Purchases.cs
aboedo: good idea. I'd started with `iOS` since we don't support macOS for Unity, but consistency with the dashboard is a great argument

RevenueCat/Scripts/Purchases.cs
aboedo: good call

RevenueCat/Scripts/Purchases.cs
aboedo: I believe Unity doesn't support Catalyst. 
@vegaro do you recall if there's a strong reason we don't support macOS? From our perspective, it shouldn't really matter if you're running macOS or not, right?

aboedo: happy to update this line to also include macOS, since it can't hurt

RevenueCat/Scripts/Purchases.cs
aboedo: I don't think we should tell folks to call setup again! That might be reeeeally outdated, you should be calling `reset` in old SDKs and `logOut` in newer ones, but never setup. Great catch!

RevenueCat/Scripts/Purchases.cs
jefago: 💯 on `Google` because we support other Android stores

RevenueCat/Scripts/Purchases.cs
vegaro: I explored it in the past and we can't reuse our current iOS code. I actually started writing the code to support it and hit a wall that I don't remember so I reverted the changes. This was back when I was by myself and didn't have much Apple platforms experience. https://docs.unity3d.com/Manual/PluginsForDesktop.html

RevenueCat/Scripts/Purchases.cs
aboedo: updated!

PR #79, opened 2022-01-10
Add BillingClient versions to VERSIONS.md

vegaro:
Just wanted to add some clarity on Unity IAP versions compatibility so decided to add it to VERSIONS.md. Let's see what you think and then I can add it to purchases-android too. Unity is the one that looks like has more compatibility issues that's why I decided to start with it.


aboedo (approved):
❤️ it. Maybe we should link to this in the readme?

vegaro:
Added to README.md in https://github.com/RevenueCat/purchases-unity/pull/79/commits/388fb3919fca38e24984fbe5491b5c1ef9157fe7

PR #81, opened 2022-01-27
Add auto sync purchases option [One off release]

vegaro:
We need to make a release with changes from https://github.com/RevenueCat/purchases-android/pull/451 and https://github.com/RevenueCat/purchases-ios/pull/1219 on top of Unity 3.2.0.

We don't want to make a public release of this until we are capable of making releases of the latest unity version with billing client 3 support. We need https://github.com/RevenueCat/purchases-android/pull/444 for that

I updated the iOS dependency to the latest v3 release (which includes https://github.com/RevenueCat/purchases-ios/pull/1219) and will build https://github.com/RevenueCat/purchases-android/pull/461 locally and include the `aar` files in the unitypackage. The billing client aar won't be included because they are in observer mode.


RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
```diff
@@ -58,17 +58,23 @@ - (void)setupPurchases:(NSString *)apiKey
              appUserID:(nullable NSString *)appUserID
             gameObject:(NSString *)gameObject
           observerMode:(BOOL)observerMode
- userDefaultsSuiteName:(nullable NSString *)userDefaultsSuiteName {
-    [[RCPurchases sharedPurchases] setDelegate:nil];
+ userDefaultsSuiteName:(nullable NSString *)userDefaultsSuiteName
+     dangerousSettings:(NSDictionary *)dangerousSettingsAsDictionary {
+    if (RCPurchases.isConfigured) {
+        [[RCPurchases sharedPurchases] setDelegate:nil];
+    }
```
vegaro, line 65: This was a tricky one to realize. The newer version of iOS throws an exception if `sharedPurchases` is accessed before configuring.

RevenueCat/Scripts/Purchases.cs
```diff
@@ -54,6 +54,16 @@ public partial class Purchases : MonoBehaviour
...
+    
+	[Header("Dangerous Settings")]
+    [Tooltip("Disable or enable subscribing to the StoreKit queue." +  
+	"If this is disabled, RevenueCat won't observe the StoreKit queue or check current purchasers, " +
+	"and it will not sync any purchase automatically. Call syncPurchases whenever a new transaction is " +
+	"completed so the receipt is sent to RevenueCat's backend. " +
+	"In iOS, consumables disappear from the receipt after the transaction is finished, so make sure purchases " +
+	"are synced before finishing any consumable transaction, otherwise RevenueCat won't register the purchase. " +
+	"Auto syncing of purchases is enabled by default.")]
+    public bool autoSyncPurchases = true;
```
vegaro, line 66: This version of Unity (3.2.0) doesn't have a runtime setup option. That is getting added with the Amazon support. Since this is a build I am building for a customer, I added it as an option in the Editor for now. 

When we release a Unity version with Amazon support this should not be available as an Editor option since it's quite a dangerous setting.

RevenueCat/Plugins/Editor/RevenueCatDependencies.xml
```diff
@@ -1,10 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 <dependencies>
-    <androidPackages>
+    <!-- <androidPackages>
```
vegaro, line 3: To avoid having to do an Android release, and since the developers are not using EDM4U in their Android builds, I decided to comment this out and to include modified `.aar` files of purchases-android built from this branch https://github.com/RevenueCat/purchases-android/pull/461

vegaro:
[Purchases.unitypackage.zip](https://github.com/RevenueCat/purchases-unity/files/7961958/Purchases.unitypackage.zip)

vegaro:
Closing since this has been sent to the developers

PR #93, opened 2022-02-23
Detect Android emulator for Android setup

### LINKED ISSUE #89: Crash when not setting `deprecatedLegacyRevenueCatAPIKey` on simulator
On 3.5.0 and 3.5.1 versions of purchases-unity, if you test using the simulator, the app will crash with a confusing message saying "Unity error: "NullReferenceException: Object reference not set to an instance of an object". 

The issue stems [from this logic in `setup`](https://github.com/RevenueCat/purchases-unity/blob/main/RevenueCat/Scripts/Purchases.cs#L105), which checks for iOS and Android physical devices, but apparently simulators don't pass the checks.

joshdholtz:
## Motivation
Fixes #89
Some Android emulators were not being detected by the `RuntimePlatform.Android` check

## Description
- Some emulators (not all) are not detected by `Application.platform == RuntimePlatform.Android`
- Adds `IsAndroidEmulator()` function
  - Attempts to initialize an `android.os.Build` class to extract the `FINGERPRINT`
  - If `FINGERPRINT` is `generic` then it should be an Android emulator
  - Wrapped in try/catch because creating `AndroidJavaClass` will fail on iOS

⁉️ Additional thought... We could just get rid of the `FINGERPRINT` and rely on `AndroidJavaClass` because we don't care if its an emulator or not. We only care about it its an Android device 🤷‍♂️

### DIFF
--- a/RevenueCat/Scripts/Purchases.cs
+++ b/RevenueCat/Scripts/Purchases.cs
@@ -115,13 +115,12 @@ public partial class Purchases : MonoBehaviour
         _wrapper.Setup(gameObject.name, apiKey, newUserId, observerMode, userDefaultsSuiteName);
     }
 
-    public bool IsAndroidEmulator()
+    private bool IsAndroidEmulator()
     {
         try
         {
             // From https://stackoverflow.com/questions/51880866/detect-if-game-running-in-android-emulator
-            AndroidJavaClass osBuild;
-            osBuild = new AndroidJavaClass("android.os.Build");
+            AndroidJavaClass osBuild = new AndroidJavaClass("android.os.Build");
             string fingerPrint = osBuild.GetStatic<string>("FINGERPRINT");
             return fingerPrint.Contains("generic");
         }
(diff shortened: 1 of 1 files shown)


RevenueCat/Scripts/Purchases.cs
```diff
@@ -114,6 +115,23 @@ private void Setup(string newUserId)
         _wrapper.Setup(gameObject.name, apiKey, newUserId, observerMode, userDefaultsSuiteName);
     }
 
+    public bool IsAndroidEmulator()
+    {
+        try
+        {
+            // From https://stackoverflow.com/questions/51880866/detect-if-game-running-in-android-emulator
+            AndroidJavaClass osBuild;
+            osBuild = new AndroidJavaClass("android.os.Build");
```
vegaro, line 124: Can we join it with the declaration of the variable?

RevenueCat/Scripts/Purchases.cs
```diff
@@ -114,6 +115,23 @@ private void Setup(string newUserId)
         _wrapper.Setup(gameObject.name, apiKey, newUserId, observerMode, userDefaultsSuiteName);
     }
 
+    public bool IsAndroidEmulator()
```
vegaro, line 118: This can be private right?

joshdholtz: Oh yeah, I meant to change that 🙈

vegaro (approved):

PR #108, opened 2022-05-30
Update play services resolver

aboedo:
Updated the Google Play Services Resolver to the latest version. 

The biggest motivator was that I was getting [this error constantly in the editor](https://support.unity.com/hc/en-us/articles/4408586248340-Why-do-I-get-Error-Ambiguous-Match-Found-Non-Mediation-PSR-in-use-). 

Tested on Android and on iOS


Subtester/Assets/Plugins/Android/gradleTemplate.properties
```diff
@@ -0,0 +1,9 @@
+org.gradle.jvmargs=-Xmx**JVM_HEAP_SIZE**M
```
aboedo, line 1: This is now needed for the resolver to work correctly - we need to have a template for `gradle.properties`, just like we have one for `mainTemplate.gradle` -> `build.gradle`.

aboedo: It also means we no longer have to set up the `android.useAndroidX=true` property manually every time, though

Subtester/Assets/Plugins/Android/mainTemplate.gradle
```diff
@@ -1,20 +1,13 @@
...
-}
 ([rootProject] + (rootProject.subprojects as List)).each { project ->
     project.repositories {
         def unityProjectPath = $/file:///**DIR_UNITYPROJECT**/$.replace("\\", "/")
         maven {
             url "https://maven.google.com"
         }
         mavenLocal()
-        jcenter()
         mavenCentral()
```
aboedo, line 11: this stuff comes from the resolver automatically

Subtester/ProjectSettings/AndroidResolverDependencies.xml
```diff
@@ -1,14 +1,15 @@
 <dependencies>
   <packages>
     <package>androidx.annotation:annotation:[1.2.0]</package>
-    <package>com.revenuecat.purchases:purchases-hybrid-common:[1.10.0]</package>
+    <package>com.revenuecat.purchases:purchases-hybrid-common:[3.0.0]</package>
```
aboedo, line 4: no entirely sure why this hadn't been updated before, I might have just missed committing it

PR #112, opened 2022-06-17
API cleanup

aboedo:
This cleans up our API, by following C# naming standards and also making numerous properties read-only. 

We might also want to kill attributionNetwork but I'll work on that as a separate PR.

### DIFF
--- a/RevenueCat/Scripts/Purchases.cs
+++ b/RevenueCat/Scripts/Purchases.cs
@@ -46,36 +46,36 @@ public partial class Purchases : MonoBehaviour
              "This property is obsolete. " +
              "Use revenueCatAPIKeyApple and revenueCatAPIKeyGoogle instead.")]
     // ReSharper disable once InconsistentNaming
-    public string deprecatedLegacyRevenueCatAPIKey;
+    public string DeprecatedLegacyRevenueCatAPIKey;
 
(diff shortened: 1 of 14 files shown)


RevenueCat/Scripts/Purchases.cs
```diff
@@ -48,36 +46,36 @@ public partial class Purchases : MonoBehaviour
...
 
     [Tooltip("RevenueCat API Key specifically for Apple platforms. Get from https://app.revenuecat.com/")]
     // ReSharper disable once InconsistentNaming
-    public string revenueCatAPIKeyApple;
+    public string RevenueCatAPIKeyApple;
 
     [Tooltip("RevenueCat API Key specifically for Android. Get from https://app.revenuecat.com/")]
     // ReSharper disable once InconsistentNaming
-    public string revenueCatAPIKeyGoogle;
+    public string RevenueCatAPIKeyGoogle;
```
aboedo, line 57: turns out that we need to revert these 🤦 I'm doing it on the #114 PR. 
Unity's naming standards for stuff that shows up in the Editor say to use camelCase, as opposed to typical C# would would be Pascal.

aboedo:
Also closing this one in favor of #114

PR #122, opened 2022-07-15
Update AMAZON-INSTRUCTIONS.md

vegaro:
I tested an observer mode installation and updated the Amazon instructions since some things were not correct. 

We still need to fix [CSDK-335](https://revenuecats.atlassian.net/browse/CSDK-335) so excluding the billingclient is not needed anymore.


vegaro:
[This PR](https://github.com/RevenueCat/purchases-android/pull/566) will remove the need for the mainTemplate.gradle extra steps

PR #125, opened 2022-07-26
rename `Setup` to `Configure`

aboedo:
Renamed `Setup` to `Configure`. 

I added a deprecation to `Setup` although it was never technically public in a non-prerelease build, because it doesn't hurt to have it and it'll make it slightly easier for anyone who was using a pre-release build to update.

### DIFF
--- a/RevenueCat/Scripts/Purchases.cs
+++ b/RevenueCat/Scripts/Purchases.cs
@@ -44,18 +44,22 @@ public partial class Purchases : MonoBehaviour
     [Tooltip("Activate if you plan to call Purchases.Configure or Purchases.Setup programmatically.")]
     public bool useRuntimeSetup;
 
-    [Tooltip("RevenueCat API Key specifically for Apple platforms. Get from https://app.revenuecat.com/")]
+    [Tooltip("RevenueCat API Key specifically for Apple platforms. Get from https://app.revenuecat.com/ " +
+             "Note: This value not be used if Purchases.useRuntimeSetup is true.")]
     public string revenueCatAPIKeyApple;
 
-    [Tooltip("RevenueCat API Key specifically for Google Play. Get from https://app.revenuecat.com/")]
+    [Tooltip("RevenueCat API Key specifically for Google Play. Get from https://app.revenuecat.com/ " +
+             "Note: This value not be used if Purchases.useRuntimeSetup is true.")]
     public string revenueCatAPIKeyGoogle;
 
     [Header("Alternative Stores")]
-    [Tooltip("RevenueCat API Key specifically for Amazon Appstore. Get from https://app.revenuecat.com/")]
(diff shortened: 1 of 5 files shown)


RevenueCat/Scripts/Purchases.cs
```diff
@@ -41,56 +41,76 @@ public partial class Purchases : MonoBehaviour
     /// </summary>
     public delegate void GetPromotionalOfferFunc(PromotionalOffer promotionalOffer, Error error);
 
-    [Tooltip("Activate if you plan to call Purchases.Setup programmatically.")]
+    [Tooltip("Activate if you plan to call Purchases.Configure or Purchases.Setup programmatically.")]
```
aboedo, line 44: did a decent amount of cleanup on these tooltips so they look a bit better

aboedo: mostly line breaks both in C# by splitting the string and in the tooltip itself with `\n`

RevenueCat/Scripts/Purchases.cs
```diff
@@ -134,10 +154,17 @@ private void Setup(string newUserId)
             .SetUseAmazon(useAmazon)
             .SetDangerousSettings(dangerousSettings);
 
-        Setup(builder.Build());
+        Configure(builder.Build());
     }
 
+    [Obsolete("Deprecated, use Configure instead.", false)]
```
aboedo, line 160: this method was never public in a final release, but it doesn't hurt to have the deprecation for anyone using a release candidate

RevenueCat/Scripts/PurchasesConfiguration.cs
```diff
@@ -41,6 +52,7 @@ public static Builder Init(string apiKey)
 
             public PurchasesConfiguration Build()
             {
+                _dangerousSettings ??= new DangerousSettings(false);
```
aboedo, line 55: without this, creating a build but not setting dangerous settings would crash

tonidero: I guess we don't have tests for this right?

RevenueCat/Scripts/Purchases.cs
tonidero: Well... I guess the only downside is that we will have an extra deprecated function to test/remove at some point... But then again, I think what you said makes sense so I'm ok with this 👍

tonidero (approved):
Newbie here but LGTM!

vegaro (approved):

RevenueCat/Scripts/PurchasesConfiguration.cs
aboedo: we don't have unit tests for anything in Unity, sadly. Hoping to get that started soon. We do have API tests, as well as integration tests that export a `.unityPackage`, import it into an iOS and an Android app, and try building (and creating an archive on iOS)

PR #128, opened 2022-07-27
added examples and better docs to new configure method

aboedo:
Improves the docstring so usage is hopefully easier to understand. 

Here's an example of how this looks in the editor: 

Screen Shot 2022-07-27 at 11 07 22 AM

### DIFF
--- a/RevenueCat/Scripts/Purchases.cs
+++ b/RevenueCat/Scripts/Purchases.cs
@@ -164,6 +164,28 @@ public partial class Purchases : MonoBehaviour
     }
     
     // ReSharper disable once MemberCanBePrivate.Global
+    /// <summary>
+    /// Use this method to configure the SDK programmatically.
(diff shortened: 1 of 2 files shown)


RevenueCat/Scripts/Purchases.cs
```diff
@@ -164,6 +164,28 @@ public void Setup(PurchasesConfiguration purchasesConfiguration)
     }
     
     // ReSharper disable once MemberCanBePrivate.Global
+    /// <summary>
+    /// Use this method to configure the SDK programmatically.
```
tonidero, line 168: Maybe mention that this should be called before any other SDK methods to work properly?

tonidero (approved):

RevenueCat/Scripts/Purchases.cs
aboedo: Great call! I'll do that in a separate PR to avoid having to rebase all of these

PR #129, opened 2022-07-27
Fix: crash when cancelling

aboedo:
There's a crash when cancelling purchases, due to a difference in behavior between SK1 and SK2. 

Ideally we'd improve behavior to unify between them, but in the meantime, this SDK has been thoroughly tested with SK1, not with SK2, so using SK1 is safer.


beylmk (approved):
👍

PR #132, opened 2022-07-29
Improve docs part 3

aboedo:
Part 3 of improving docs. 

Addresses comments from PR reviews in parts 1 and 2

### DIFF
--- a/RevenueCat/Scripts/Purchases.cs
+++ b/RevenueCat/Scripts/Purchases.cs
@@ -373,9 +373,11 @@ public partial class Purchases : MonoBehaviour
     /// You shouldn't use this method if you have your own account system. In that case "restoration" is provided
     /// by your app passing the same <c>appUserID</c> used to purchase originally.
     ///
-    /// - Note: This may force your users to enter their Store password so should only be performed on request of
+    /// <remarks>- Note: This may force your users to enter their Store password so should only be performed on request of
     /// the user. Typically with a button in settings or near your purchase UI. Use
     /// <see cref="SyncPurchases"/> if you need to restore transactions programmatically.
+    /// </remarks>
+    /// <seealso href="https://docs.revenuecat.com/docs/restoring-purchases"/>
     ///
     /// <param name="callback"> A <see cref="CustomerInfoFunc"/> which will contain a <see cref="CustomerInfo"/>
     /// if restoration was successful, or an error otherwise. </param>
@@ -575,6 +577,7 @@ public partial class Purchases : MonoBehaviour
     /// on the device does not contain subscriptions, but the user has made subscription purchases, this method
     /// won't be able to restore them. Use <see cref="RestorePurchases"/> to cover those cases.
     /// </remarks>
+    /// <seealso href="https://docs.revenuecat.com/docs/restoring-purchases"/>
     ///
     public void SyncPurchases()
     {
--- a/RevenueCat/Scripts/Purchases.cs
+++ b/RevenueCat/Scripts/Purchases.cs
@@ -136,12 +136,18 @@ public partial class Purchases : MonoBehaviour
     /// Use this method to configure the SDK programmatically.
     /// To use this, you *must* check <c>useRuntimeSetup</c> in the Unity IDE UI. 
     /// The values used through this setup will override values set through the Unity IDE UI.
+    /// You should call this method as soon as possible in your app's lifecycle, and before any other calls to the SDK.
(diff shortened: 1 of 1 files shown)


RevenueCat/Scripts/Purchases.cs
```diff
@@ -411,8 +396,41 @@ public void RestorePurchases(CustomerInfoFunc callback)
...
+    /// <summary>
+    /// Callback function for <see cref="Purchases.LogIn"/>. 
+    /// </summary>
+    /// <param name="customerInfo"> The <see cref="CustomerInfo"/> of the user if the request was successful.
+    /// Null otherwise.</param>
+    /// <param name="created"> True if the user was registered in RevenueCat's database for the first time upon
+    /// making this call. False if a user with this <c>appUserID</c> already existed in RevenueCat's database for
+    /// this app.</param>
+    /// <param name="error"> If the request was unsuccessful, contains an error. Null otherwise. </param>
+    public delegate void LogInFunc(CustomerInfo customerInfo, bool created, Error error);
```
tonidero, line 408: This was the same as the previous name, but I wonder if this would be better named `LogInDelegate` or `LogInCallback` (Not familiar with Unity or C# so this might be the style here). Not for this PR in any case.

RevenueCat/Scripts/Purchases.cs
```diff
@@ -720,6 +933,17 @@ public void CollectDeviceIdentifiers()
...
         _wrapper.CollectDeviceIdentifiers();
     }
+    
+    /// <summary>
+    /// Callback function containing the result of CanMakePayments
+    /// <param name="canMakePayments">A bool value indicating whether billing
+    /// is supported for the current user (meaning IN-APP purchases are supported),
+    /// and, if provided, whether a list of specified BillingFeatures are supported.
+    /// This will be false if there is an error</param>
+    /// <param name="error">An Error object or null if successful.</param>
```
tonidero, line 943: I've been wondering why sometimes we are adding the `<param>` inside the `<summary>` tag and sometimes outside. Does it organize the docs differently?

RevenueCat/Scripts/Purchases.cs
```diff
@@ -499,40 +609,106 @@ public void SyncPurchases()
...
+    /// Invalidates the cache for customer information.
+    /// </summary>
+    /// 
+    /// <remarks>Most apps will not need to use this method; invalidating the cache can leave your app in an invalid state.
+    /// Refer to https://docs.revenuecat.com/docs/purchaserinfo#section-get-user-information
+    /// for more information on using the cache properly.
+    /// </remarks>
+    ///
+    /// This is useful for cases where customer information might have been updated outside of the app, like if a
+    /// promotional subscription is granted through the RevenueCat dashboard.
```
tonidero, line 672: Will this text outside the `<summary>` tag be included in the docs?

tonidero (approved):
Just some questions but looks good in general

RevenueCat/Scripts/Purchases.cs
aboedo: it should always be outside

RevenueCat/Scripts/Purchases.cs
aboedo: yeah, it'll still be there. I have yet to see how it looks when you generate to a website, but for now in the IDE it does work. My hope is that it'll render into a separate section, like in iOS, where you have the summary and the "discussion" as separate things ([example here](https://revenuecat.github.io/purchases-ios-docs/4.9.0/documentation/revenuecat/purchases/configure(with:)-6oipy))

RevenueCat/Scripts/Purchases.cs
aboedo: did a pass through this file moving all params out of summary, thanks for calling this out!!

RevenueCat/Scripts/Purchases.cs
aboedo: The Unity SDK needs to have a _type_ for the callback, and it also needs to have a variable that stores a callback that was passed in, to call it once the method is finished. 

Our naming scheme right now is: 
The _type_ of callback for the method is defined as `FooFunc`, and then the variable is `FooCallback`. 

Full disclosure: I've no idea why we follow this naming scheme. And / or whether that's even remotely standard.

aboedo: I'll ask our more knowledgeable C# devs

PR #134, opened 2022-08-01
Update amazon instructions

aboedo:
Updates instructions for Amazon to include how to use it in observer mode, now that it's supported.

### DIFF
--- a/AMAZON-INSTRUCTIONS.md
+++ b/AMAZON-INSTRUCTIONS.md
@@ -67,7 +67,3 @@ public PurchaseProcessingResult ProcessPurchase(PurchaseEventArgs e)
     return PurchaseProcessingResult.Complete;
 }
  ```
-
- - Perform a Resolve using the editor option `Assets/External Dependency Manager/Android Resolver/Resolve` menu. This will add the right dependencies to the `mainTemplate.gradle`.
-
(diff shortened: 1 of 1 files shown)


AMAZON-INSTRUCTIONS.md
```diff
@@ -18,12 +18,56 @@ This release adds pre-release support for Amazon store.
...
+        Decimal.ToDouble(e.purchasedProduct.metadata.localizedPrice)
+    );
+    return PurchaseProcessingResult.Complete;
+}
+ ```
+
+ - Perform a Resolve using the editor option `Assets/External Dependency Manager/Android Resolver/Resolve` menu. This will add the right dependencies to the `mainTemplate.gradle`.
 
-RevenueCat doesn't support Amazon in observer mode yet
+ - In observer mode, adding the Amazon in-app-purchasing library jar is not necessary since it will be added by Unity IAP when targeting the Amazon Store
```
vegaro, line 73: I would remove this step too. In the early versions of our Amazon support, the instructions included a reference to adding the amazon jar, but that's not needed anymore because the Amazon dependency is downloaded from Maven.

aboedo:
@vegaro updates done!

vegaro (approved):

PR #140, opened 2022-08-25
Automate release process with internal fastlane plugin

tonidero:
This PR uses fastlane to automate several tasks related to new releases of the unity sdk. 

### TODO
- [x] Update CircleCI envs to add github tokens and github user/email.


VERSIONS.md
```diff
@@ -1,32 +1,32 @@
-| Version | iOS version | Android version | Common version | BillingClient version | Unity IAP compatibility  |
```
tonidero, line 1: I removed the `BillingClient version` column here... I think it's not that useful and helps simplify things a bit.

fastlane/Fastfile
```diff
@@ -0,0 +1,138 @@
...
+    versions_file_path: versions_file_path,
+    new_sdk_version: current_version_number,
+    hybrid_common_version: parse_previous_hybrid_version
+  )
+  append_previous_version_compatibility_to_versions_file(versions_file_path)
+  commit_current_changes(commit_message: 'Update VERSIONS.md')
+  push_to_git_remote(set_upstream: true)
+end
+
+def append_previous_version_compatibility_to_versions_file(versions_file_path)
```
tonidero, line 114: Honestly this might not be the cleanest... I thought about changing the `update_hybrids_versions_file` action to return the string instead and move part of the logic to the SDKs... But I thought this was not too bad. Lmk if you think I should change it.

VERSIONS.md
aboedo: It's _somewhat_ useful for folks using observer mode, but the vast majority of observer mode devs in Unity will be using Unity IAP, so as long as we still have that one this seems okay.

.circleci/config.yml
```diff
@@ -124,17 +131,46 @@ commands:
...
       - save_cache:
           key: gem-cache-{{ checksum "<< parameters.directory >>/Gemfile.lock" }}
           paths:
             - vendor/bundle
-  
+
+  set_up_git_user:
+    description: Set git config parameters
+    steps:
+      - run: git config --global user.email "dev+ci@revenuecat.com"
```
aboedo, line 148: not needed for this PR, but we should probably move this stuff to env variables

fastlane/Fastfile
```diff
@@ -0,0 +1,138 @@
...
+  Dir.chdir(get_root_folder) do
+    File.open(versions_file_path, 'r+') do |file|
+      lines = file.each_line.to_a
+      previous_to_last_version_compatibility = lines[3].split('|')[-2].strip
```
aboedo, line 118: I think it'd be useful to include a comment here to show what the format looks like, i.e.: 
```ruby
# format:
| Version | iOS version | Android version | Common version | Unity IAP compatibility  |
| 4.1.0   | 4.10.1      | 5.4.1           | 4.1.2          | Unity IAP 3.3.0+         |
```

fastlane/Fastfile
```diff
@@ -0,0 +1,138 @@
...
+      lines[2] = lines[2].strip + " #{previous_to_last_version_compatibility} |\n"
+      file.rewind
+      file.write(lines.join)
+    end
+  end
+end
+
+def parse_previous_hybrid_version
+  Dir.chdir(get_root_folder) do
+    return (sh "cat RevenueCat/Plugins/Editor/RevenueCatDependencies.xml | grep '<iosPod name=\"PurchasesHybridCommon\" version=' | awk '{print($3)}'  | sed \"s/version=//g;s/\\\"//g\"").strip
```
aboedo, line 128: could we do this in ruby? 
otherwise, could we make this a multiline string for readability? i.e.: 
```ruby
    return (sh "cat RevenueCat/Plugins/Editor/RevenueCatDependencies.xml"
      + " | grep '<iosPod name=\"PurchasesHybridCommon\" version='"
      + " | awk '{print($3)}'"
      + " | sed \"s/version=//g;s/\\\"//g\"")
      .strip
```

.circleci/config.yml
tonidero: Actually, I will probably change it for this PR, since it should be done like that yeah.

VERSIONS.md
vegaro: I agree with @aboedo 

I don't think anyone is using BillingClient directly when in observer mode

.circleci/config.yml
vegaro: And this whole command can be removed if using our orb 👍

aboedo (approved):

.circleci/config.yml
```diff
@@ -241,23 +248,63 @@ jobs:
           working_directory: IntegrationTests/buildiOS
           command: bundle exec fastlane archive workspace:"buildiOS/Unity-iPhone.xcworkspace"
 
+  prepare-next-version:
```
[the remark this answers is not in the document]
tonidero: Agreed! There are a few jobs that would be good to move to the orb... Main problem is that some of them have slight differences... For example, in iOS this is executed with `bundle exec fastlane ios prepare_next_version`... But we can try to unify them 👍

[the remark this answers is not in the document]
aboedo: I'd argue that it's probably a _good_ thing for us to be forced to unify as much as we can, so a future dev doesn't need to wonder why we have a slightly different command in iOS than Unity for the same thing

PR #142, opened 2022-08-26
Subtester: Fix unity android export

aboedo:
Exporting SubTester for Android was getting pretty annoying: 

Process before: 
- Export project
- Go to project, update gradle version to 7.0.2
- Go to gradle.properties and remove the `enableR8` line
- Go to build.gradle for the project and update the compile sdk to 31
- Go to the android manifest for the unity framework and add android:exported
- Done! Ready to run. 

Process after: 
- Export project
- Select Cancel when you see this annoying message that I couldn't find a way around from Unity. Clicking cancel will still export the project. 
image
- Done! Ready to run. 

Ideas and suggestions for the last prompt from Unity are welcome. 

Also, this fixes a bug in the purchase buttons where they might show up off screen because I refactored code and forgot to re-test 🤦

### DIFF
--- a/Subtester/Assets/Plugins/Android/baseProjectTemplate.gradle
+++ b/Subtester/Assets/Plugins/Android/baseProjectTemplate.gradle
@@ -1,0 +1,29 @@
+allprojects {
+    buildscript {
+        repositories {**ARTIFACTORYREPOSITORY**
+            google()
+            jcenter()
+        }
+
+        dependencies {
+            // If you are changing the Android Gradle Plugin version, make sure it is compatible with the Gradle version preinstalled with Unity
+            // See which Gradle version is preinstalled with Unity here https://docs.unity3d.com/Manual/android-gradle-overview.html
+            // See official Gradle and Android Gradle Plugin compatibility table here https://developer.android.com/studio/releases/gradle-plugin#updating-gradle
+            // To specify a custom Gradle version in Unity, go do "Preferences > External Tools", uncheck "Gradle Installed with Unity (recommended)" and specify a path to a custom Gradle version
+            classpath 'com.android.tools.build:gradle:7.0.2'
+            **BUILD_SCRIPT_DEPS**
+        }
+    }
+
+    repositories {**ARTIFACTORYREPOSITORY**
+        google()
+        jcenter()
+        flatDir {
+            dirs "${project(':unityLibrary').projectDir}/libs"
+        }
+    }
+}
+
+task clean(type: Delete) {
+    delete rootProject.buildDir
+}
--- a/Subtester/Assets/Plugins/Android/gradleTemplate.properties
+++ b/Subtester/Assets/Plugins/Android/gradleTemplate.properties
@@ -1,6 +1,5 @@
 org.gradle.jvmargs=-Xmx**JVM_HEAP_SIZE**M
 org.gradle.parallel=true
-android.enableR8=**MINIFY_WITH_R_EIGHT**
 unityStreamingAssets=**STREAMING_ASSETS**
 # Android Resolver Properties Start
 android.useAndroidX=true
--- a/Subtester/ProjectSettings/ProjectSettings.asset
+++ b/Subtester/ProjectSettings/ProjectSettings.asset
@@ -163,7 +163,7 @@ PlayerSettings:
     tvOS: 0
   overrideDefaultApplicationIdentifier: 0
   AndroidBundleVersionCode: 97
-  AndroidMinSdkVersion: 29
+  AndroidMinSdkVersion: 22
   AndroidTargetSdkVersion: 31
   AndroidPreferredInstallLocation: 1
   aotOptions: 
--- a/Subtester/ProjectSettings/ProjectSettings.asset
+++ b/Subtester/ProjectSettings/ProjectSettings.asset
@@ -134,7 +134,7 @@ PlayerSettings:
     16:10: 1
     16:9: 1
     Others: 1
-  bundleVersion: 0.1
+  bundleVersion: 97
   preloadedAssets: []
   metroInputSource: 0
   wsaTransparentSwapchain: 0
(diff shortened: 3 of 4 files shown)


aboedo:
This last prompt happens because apparently Unity's looking for you to use `enableR8`, which replaces proguard. 
If you accept, Unity adds the `enableR8` line, and then compilation fails because that property is obsoleted.

So we'd need to convince Unity not to look for that line in the first place, but I'm not sure how to do that. 
Still, this is an improvement, and even if you accidentally accept all you need to do afterwards is remove the property.

RevenueCat/Example/PurchasesListener.cs
```diff
@@ -59,15 +65,13 @@ private void Start()
             else
             {
                 Debug.Log("offerings received " + offerings.ToString());
-                var yOffset = 0;
```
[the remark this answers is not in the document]
aboedo: oops, this was from the previous PR 😂

Subtester/Assets/Plugins/Android/baseProjectTemplate.gradle
```diff
@@ -0,0 +1,29 @@
...
+            google()
+            jcenter()
+        }
+
+        dependencies {
+            // If you are changing the Android Gradle Plugin version, make sure it is compatible with the Gradle version preinstalled with Unity
+            // See which Gradle version is preinstalled with Unity here https://docs.unity3d.com/Manual/android-gradle-overview.html
+            // See official Gradle and Android Gradle Plugin compatibility table here https://developer.android.com/studio/releases/gradle-plugin#updating-gradle
+            // To specify a custom Gradle version in Unity, go do "Preferences > External Tools", uncheck "Gradle Installed with Unity (recommended)" and specify a path to a custom Gradle version
+            classpath 'com.android.tools.build:gradle:7.0.2'
```
aboedo, line 13: gradle plugin version specified here

Subtester/ProjectSettings/ProjectSettings.asset
```diff
@@ -134,7 +134,7 @@ PlayerSettings:
     16:10: 1
     16:9: 1
     Others: 1
-  bundleVersion: 0.1
+  bundleVersion: 97
```
aboedo, line 137: this isn't needed, but doesn't hurt. It's just the current latest build in live app testing

Subtester/ProjectSettings/ProjectSettings.asset
```diff
@@ -160,9 +162,9 @@ PlayerSettings:
     iPhone: 0
     tvOS: 0
   overrideDefaultApplicationIdentifier: 0
-  AndroidBundleVersionCode: 1
-  AndroidMinSdkVersion: 22
-  AndroidTargetSdkVersion: 0
+  AndroidBundleVersionCode: 97
+  AndroidMinSdkVersion: 29
+  AndroidTargetSdkVersion: 31
```
aboedo, line 167: without updating to target 31 we get a compilation issue

aboedo: it seems like one of the libraries requires minsdk 31

Subtester/ProjectSettings/ProjectSettings.asset
tonidero: For Amazon only right? I think this would likely get outdated soon, considering we reuse the same app in multiple sdks for amazon... NABD though.

Subtester/ProjectSettings/ProjectSettings.asset
```diff
@@ -160,9 +162,9 @@ PlayerSettings:
     iPhone: 0
     tvOS: 0
   overrideDefaultApplicationIdentifier: 0
-  AndroidBundleVersionCode: 1
-  AndroidMinSdkVersion: 22
-  AndroidTargetSdkVersion: 0
+  AndroidBundleVersionCode: 97
+  AndroidMinSdkVersion: 29
```
tonidero, line 166: That's a pretty high minSdkVersion... Did you have issues with a lower one?

Subtester/ProjectSettings/ProjectSettings.asset
```diff
@@ -145,11 +145,13 @@ PlayerSettings:
     enable360StereoCapture: 0
   isWsaHolographicRemotingEnabled: 0
   enableFrameTimingStats: 0
+  enableOpenGLProfilerGPURecorders: 1
   useHDRDisplay: 0
   D3DHDRBitDepth: 0
   m_ColorGamuts: 00000000
   targetPixelDensity: 30
   resolutionScalingMode: 0
+  resetResolutionOnWindowResize: 0
```
tonidero, line 154: Hmm so I'm not sure I understand it but should this be enabled? Seems like something we might want in case someone resizes it... Not sure though

Subtester/Assets/Plugins/Android/gradleTemplate.properties
```diff
@@ -1,6 +1,5 @@
 org.gradle.jvmargs=-Xmx**JVM_HEAP_SIZE**M
 org.gradle.parallel=true
-android.enableR8=**MINIFY_WITH_R_EIGHT**
```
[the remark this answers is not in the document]
tonidero: Were you able to fix compilation in Unity after removing this?

tonidero (approved):
Just some questions but looks good!

Subtester/Assets/Plugins/Android/gradleTemplate.properties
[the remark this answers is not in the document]
aboedo: yeah, with this removed the Unity IDE issues a warning (and the prompt when you click on export) but things work correctly

Subtester/ProjectSettings/ProjectSettings.asset
aboedo: yeah, this was just me being lazy to roll this change back, and since it doesn't make a difference either way

Subtester/ProjectSettings/ProjectSettings.asset
aboedo: I've no idea tbh, this was just an automated change by Unity

Subtester/ProjectSettings/ProjectSettings.asset
aboedo: I can try lower ones. I suppose this would affect us if devs are using test devices with lower sdk versions, right?

Subtester/Assets/Plugins/Android/gradleTemplate.properties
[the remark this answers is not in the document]
tonidero: Yeah basically... Also, to make sure we keep the SDK compatible with lower versions of android as well.

Subtester/ProjectSettings/ProjectSettings.asset
tonidero: Yeah basically... Also, to make sure we keep the SDK compatible with lower versions of android as well.

aboedo:
@tonidero merging for now, but let me know if there's anything else I should cover in a follow-up!

PR #150, opened 2022-09-07
Move to SNAPSHOT version on every commit to main if needed

tonidero:
### Description
This PR updates the fastlane plugin to include https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/31. With this, every time we commit to `main`, if the version is not a SNAPSHOT, it will create a PR updating to the next SNAPSHOT version. Additionally, I moved this job to a different workflow to keep the `deploy` workflow a bit more tidy.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,13 @@ aliases:
       branches:
         ignore: /.*/
 
+  only-main-branch: &only-main-branch
+    filters:
+      tags:
+        ignore: /.*/
+      branches:
+        only: main
+
 orbs:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -306,7 +313,11 @@ workflows:
           <<: *release-tags
           requires:
             - export-package
+
+  snapshot-bump:
+    when:
+      not:
+        equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
```
tonidero, line 320: This isn't really needed just yet here (until we have added release trains) but I thought I would add it.

vegaro (approved):

PR #145, opened 2022-09-02
Bump PHC to 4.2.1

tonidero:
This PR bumps PHC to 4.2.1. As part of that change, we have added `TransactionIdentifier` and `ProductIdentifier` to `StoreTransaction` while deprecating the old `RevenueCatId` and `ProductId`.

This PR also contains a couple of small fixes for the release automation process.

PR #151, opened 2022-09-08
Support for `usesStoreKit2IfAvailable` config option

tonidero:
This PR adds support for `usesStoreKit2IfAvailable` config option, both from the UI and the scripts. When this option is enabled, in iOS devices it will use store kit 2 to perform purchases and related operations.

### DIFF
--- a/RevenueCat/Scripts/PurchasesWrapperAndroid.cs
+++ b/RevenueCat/Scripts/PurchasesWrapperAndroid.cs
@@ -48,7 +48,7 @@ public class PurchasesWrapperAndroid : IPurchasesWrapper
         }
     }
 
-    public void Setup(string gameObject, string apiKey, string appUserId, bool observerMode,
+    public void Setup(string gameObject, string apiKey, string appUserId, bool observerMode, bool usesStoreKit2IfAvailable,
         string userDefaultsSuiteName, bool useAmazon, string dangerousSettingsJson)
     {
         CallPurchases("setup", apiKey, appUserId, gameObject, observerMode, userDefaultsSuiteName, useAmazon,
@@ -115,7 +115,7 @@ public class PurchasesWrapperAndroid : IPurchasesWrapper
     {
         CallPurchases("syncObserverModeAmazonPurchase", productID, receiptID, amazonUserID, isoCurrencyCode, price);
     }
-    
+
     public void SetAutomaticAppleSearchAdsAttributionCollection(bool enabled)
     {
         // NOOP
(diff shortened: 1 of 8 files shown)


RevenueCat/Scripts/StoreTransaction.cs
```diff
@@ -52,19 +52,23 @@ public class StoreTransaction
         public StoreTransaction(JSONNode response)
         {
             TransactionIdentifier = response["transactionIdentifier"];
-            RevenueCatId = response["transactionIdentifier"];
             ProductIdentifier = response["productIdentifier"];
+            #pragma warning disable 618 // Disable Obsolete warning
```
tonidero, line 56: This isn't related to `usesStoreKit2IfAvailable`, but related to #145 . This just disables some warnings that were showing in the Unity UI after that change since `RevenueCatId` and `ProductId` were marked as `Obsolete`.

RevenueCat/Scripts/PurchasesWrapperAndroid.cs
```diff
@@ -48,7 +48,7 @@ public void GetProducts(string[] productIdentifiers, string type = "subs")
         }
     }
 
-    public void Setup(string gameObject, string apiKey, string appUserId, bool observerMode,
+    public void Setup(string gameObject, string apiKey, string appUserId, bool observerMode, bool usesStoreKit2IfAvailable,
```
tonidero, line 51: On android, this parameter is just ignored.

RevenueCat/Scripts/PurchasesWrapperiOS.cs
```diff
@@ -6,12 +6,13 @@
...
-    private static extern void _RCSetupPurchases(string gameObject, string apiKey, string appUserId, bool observerMode, string userDefaultsSuiteName);
+    private static extern void _RCSetupPurchases(string gameObject, string apiKey, string appUserId, bool observerMode,
+                                                 bool usesStoreKit2IfAvailable, string userDefaultsSuiteName);
 
-    public void Setup(string gameObject, string apiKey, string appUserId, bool observerMode,
+    public void Setup(string gameObject, string apiKey, string appUserId, bool observerMode, bool usesStoreKit2IfAvailable,
         string userDefaultsSuiteName, bool useAmazon, string dangerousSettingsJson)
     {
-        _RCSetupPurchases(gameObject, apiKey, appUserId, observerMode, userDefaultsSuiteName);
+        _RCSetupPurchases(gameObject, apiKey, appUserId, observerMode, usesStoreKit2IfAvailable, userDefaultsSuiteName);
```
tonidero, line 15: I didn't change it in this PR but we are not passing the `dangerousSettingsJson` parameter down to the iOS SDK it seems (though I guess we would need to parse it before). Should we do that? @aboedo.  I can tackle it in a different PR if so.

aboedo: oops
yeah, let's do it as a separate PR

aboedo (approved):

PR #152, opened 2022-09-08
Fix support for DangerousSettings in iOS

tonidero:
While working on #151 I noticed that `DangerousSettings` was not fully wired up in iOS. This fixes that so we can now pass dangerous settings to the native layer in iOS.


aboedo (approved):

aboedo:
let's get a build out tomorrow for this

PR #162, opened 2022-10-13
fix compatibility issues with C# 7.3

aboedo:
We were using a couple of features from C# 8 that aren't compatible with older versions, this updates for compatibility with 7.3

### DIFF
--- a/RevenueCat/Example/PurchasesListener.cs
+++ b/RevenueCat/Example/PurchasesListener.cs
@@ -387,7 +387,7 @@ public class PurchasesListener : Purchases.UpdatedCustomerInfoListener
             {
                 var products = GetPackages(offerings)
                     .Select(package => package.StoreProduct) // map to product ids
-                    .Where(storeProduct => storeProduct.Discounts is { Length: > 1 }) // map to product ids
+                    .Where(product => product.Discounts != null && product.Discounts.Any()) // map to product ids
                     .ToArray();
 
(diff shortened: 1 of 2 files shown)


RevenueCat/Example/PurchasesListener.cs
```diff
@@ -448,7 +448,7 @@ void BuyFirstPackageWithDiscount()
             else
             {
                 var packages = GetPackages(offerings)
-                    .Where(package => package.StoreProduct.Discounts is { Length: > 1 }) // map to product ids
+                    .Where(pkg => pkg.StoreProduct.Discounts != null && pkg.StoreProduct.Discounts.Any()) // map to product ids
```
aboedo, line 451: same exact notes as the other thing

RevenueCat/Scripts/PurchasesConfiguration.cs
```diff
@@ -81,7 +81,7 @@ public static Builder Init(string apiKey)
 
             public PurchasesConfiguration Build()
             {
-                _dangerousSettings ??= new DangerousSettings(false);
+                _dangerousSettings = _dangerousSettings ?? new DangerousSettings(false);
```
aboedo, line 84: null-coalescing assignment operator `??=` is new in c# 8.0

vegaro (approved):

tonidero (approved):

PR #175, opened 2022-11-10
Support unity package manager (UPM)

tonidero:
https://revenuecats.atlassian.net/browse/CSDK-330

In this PR, we add the necessary configuration to be able to use `purchases-unity` through UPM. With this PR, devs will be able to import the `purchases-unity` package with a git url. Some followup work (after this is merged) will be done to add this package to `openupm`.

The main changes in this PR are:
- Modified `Subtester` to use `purchases-unity` as a local package.
- Added tests to run API tests with both `Subtester` (a project using `purchases-unity` as a package) and `IntegrationTests` (a project that keeps using a `unitypackage`)
- Added `RevenueCat/package.json` with the information of the package.
- Removed `ExternalDependenciesManager` dependency downloaded in `Subtester`. Instead use the version in openupm.


.circleci/config.yml
```diff
@@ -312,22 +377,26 @@ workflows:
...
-            - build-test-ios
+            - build-integration-tests-ios
       - hold:
           type: approval
           requires:
-            - build-test-android
+            - build-integration-tests-android
+            - build-subtester-android
+            - build-subtester-ios
             - archive-ios
```
tonidero, line 400: `archive-ios` is currently only run for `IntegrationTests`. I think we don't need to run it as well for `Subtester`, since it seems the compilation of the APITests already happens in the `build-subtester-ios` job but lmk if you think we should do this as well.

RevenueCat/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "in-app purchase",
+    "subscriptions",
+    "iap",
+    "iOS",
+    "Apple",
+    "Android"
+  ],
+  "repoUrl": "https://github.com/RevenueCat/purchases-unity",
+  "dependencies": {
+    "com.google.external-dependency-manager" : "1.2.172"
```
tonidero, line 25: I believe this should automatically add this dependency when devs add purchases-unity using openupm cli. I will test it to confirm once we publish the package in openupm.

Subtester/Packages/manifest.json
```diff
@@ -1,5 +1,6 @@
 {
   "dependencies": {
+    "com.revenuecat.purchases-unity": "file:../../RevenueCat",
```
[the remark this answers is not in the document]
vegaro: Oh I think this makes things easier to understand and work with

scripts/create-unity-package.sh
```diff
@@ -10,17 +10,28 @@ done
 PROJECT="$PWD/Subtester"
 PACKAGE="$PWD/Purchases.unitypackage"
 PACKAGE_UNITY_IAP="$PWD/Purchases-UnityIAP.unitypackage"
+
+# To export a .unitypackage, we need to have the SDK files inside the Subtester project
+# Additionally, we need to remove the UPM dependency from the manifest to export it,
+# otherwise it will cause issues due to duplicated symbols.
+SYMBOLIC_LINK_PATH="$PROJECT/Assets/RevenueCat"
+MANIFEST_JSON_PATH="$PROJECT/Packages/manifest.json"
+ln -s "$PWD/RevenueCat" "$PROJECT/Assets/"
```
vegaro, line 19: So this is creating a link between the `RevenueCat` folder in the root and `Subtester/Assets`. Why?

scripts/create-unity-package.sh
```diff
@@ -10,17 +10,28 @@ done
...
+awk '!/com.revenuecat.purchases-unity/' $MANIFEST_JSON_PATH > temp && mv temp $MANIFEST_JSON_PATH
```
vegaro, line 20: This is removing the UPM dependency correct?

scripts/create-unity-package.sh
```diff
@@ -10,17 +10,28 @@ done
...
+
 FOLDERS_TO_EXPORT=$(cd $PROJECT; find Assets/RevenueCat/* Assets/PlayServicesResolver Assets/ExternalDependencyManager -type d -prune)
```
vegaro, line 22: `Subtester/Assets/RevenueCat` doesn't exist anymore correct? How is this finding the plugin files?

scripts/create-unity-package.sh
tonidero: Yes. If we try to export a `.unitypackage` from Subtester it won't export the RevenueCat SDK files even though they are part of the project using a local package. The only way to export a `unitypackage` with these assets is to have these files inside the `Assets` folder of the project. This is only done as part of the process to create the `unitypackage`.

scripts/create-unity-package.sh
tonidero: Yes, probably good to add a comment here explaining it. But basically, if we leave the UPM dependency here and we add the symlink to the RevenueCat folder, it will fail to create the `unitypackage` due to duplicate files, so we need to remove the UPM dependency as part of creating the `unitypackage`. Will add a comment explaining it.

scripts/create-unity-package.sh
tonidero: It does after adding the symlink above.

scripts/create-unity-package.sh
tonidero: Oh actually, I had a comment above, I forgot 😅. Do you think we should make it clearer here?

scripts/create-unity-package.sh
aboedo: definitely worth commenting thoroughly. 
Furthermore... this script just keeps getting more and more complicated over time. We've been putting it off for a while, but it feels like we might have already crossed the threshold where it should be moved to ruby for clarity and maintainability. 

We can do that as a separate task, but it would definitely be easier with this fresh in our memories than when we need another maintainer needs to actually change anything here.

scripts/create-unity-package.sh
aboedo: yeah, default to over-explaining for stuff like this, since this is not a trivial line. a maintainer first needs a min to understand what this even does and how, then another to figure out why. we need to save them as much time as possible.

aboedo: you know the old phrase "write code like the person maintaining it is a serial killer who knows where you live"

scripts/create-unity-package.sh
vegaro: Ahhh got it

scripts/create-unity-package.sh
vegaro: +1 for migrating it as a followup

vegaro:
Do you mind doing a comparison of the contents of the `.unitypackage` before and after this changes? To check the contents you can import the package to a project and it Unity will show you the files to import. I can also do it if you enable the `export-package` job in this branch. Also feel free to ignore if you've already compared both and made sure they contain the exact same files :)

tonidero:
It's the same:
| Last stable version unitypackage | This branch unitypackage |
| ---------- | --------- |
| image | image |

scripts/create-unity-package.sh
tonidero: Absolutely, I think moving to fastlane makes a lot of sense.

tonidero: Created Jira task for it here: https://revenuecats.atlassian.net/browse/CSDK-601

RevenueCat/Editor/RevenueCatPostInstall.cs
```diff
@@ -1,12 +1,12 @@
 
-#if UNITY_IOS
```
vegaro, line 2: Oh why this? `UNITY_EDITOR`

RevenueCat/Scripts/revenuecat.purchases-unity.asmdef
```diff
@@ -0,0 +1,14 @@
+{
```
vegaro, line 1: Do we need these asmdef both here and in RevenueCat/Editor?

RevenueCat/package.json
vegaro: It might need an extra repository but I am not sure

vegaro (approved):

RevenueCat/Scripts/revenuecat.purchases-unity.asmdef
tonidero: Hmm well, this one is needed so we can access these files in projects using purchases-unity as a package. For example, for Subtester, without this, it fails with a bunch of errors like this: `Assets/Scripts/PurchasesListener.cs(8,34): error CS0246: The type or namespace name 'Purchases' could not be found (are you missing a using directive or an assembly reference?)`.

About the Editor one, I believe it would be needed if it needs to be accessed in projects. Honestly, not sure if it's needed but I added it just in case. Do you think we should remove it?

RevenueCat/Editor/RevenueCatPostInstall.cs
tonidero: Well, without this, I was getting a bunch of compilation issues when exporting Subtester as an iOS project.
image

It seemed like it was setting the correct flags and adding storekit correctly, so it seems to be fine. Do lmk if you think this could bring issues though!

Subtester/Packages/manifest.json
```diff
@@ -1,5 +1,7 @@
...
+    "com.google.external-dependency-manager": "1.2.172",
```
tonidero, line 4: @vegaro @aboedo  I missed that after I added `com.google.external-dependency-manager` as a dependency in the `RevenueCat` package, I started getting an error in the editor due to not finding this package as a dependency in Subtester. It seemed to still be able to compile fine but in order to remove that error, I removed the version of this dependency that was checked in the repo and added it here as a package registry dependency. A couple of things to note:
- During the `unitypackage` creation script, I had to remove this dependency as well, to avoid duplicate files errors.
- We are using the latest version available in openupm (1.2.172) but the latest version of the dependency is 1.2.175. I think we should try to update to the latest soon, either by submitting the new version to the existing openupm dependency (which would require updating this repo: https://github.com/xinatcg/GoogleExternalDependencyManager) or opening our own repo to contain the externaldependencymanager as a package and we have more control over it.

RevenueCat/Editor/RevenueCatPostInstall.cs
vegaro: I think if the exported project is fine this is good

RevenueCat/Scripts/revenuecat.purchases-unity.asmdef
vegaro: I think it's good. I am actually not entirely sure what these files do lol

Subtester/Packages/manifest.json
vegaro: > During the unitypackage creation script, I had to remove this dependency as well, to avoid duplicate files errors.
We need to package it to the unitypackage though https://github.com/googlesamples/unity-jar-resolver#plugin-redistribution , are we not packaging it now inside the unitypackage? I think we need to include it for whoever is not using the unity package manager correct?

vegaro: Isn't the EDM4U plugin available in the Unity Package Manager? Can we say our plugin depends on that one? It probably requires adding another repository, I am not sure

vegaro: Ok, this has recently changed according to this post https://forum.unity.com/threads/add-external-dependency-manager-via-package-manager.1345652/ 

I knew I had installed it using the Package Manager before haha

tonidero: Right, so basically Subtester now uses the EDM4U version from openupm (the old scoped registry stopped being updated [as you said ](https://forum.unity.com/threads/add-external-dependency-manager-via-package-manager.1345652/)).

When we export the `unitypackage` though, we remove both the `purchases-unity` and `EDM4U` package from Subtester. Then, we add a symbolic link to the purchases-unity folder and download the latest version from the [unity-jar-resolver](https://github.com/googlesamples/unity-jar-resolver) repo and import it into the Subtester/Assets folder. This is quite complicated and hacky... We could create another project that is only used for exporting the `unitypackage`, which might be simpler, but as you said, on the `unity-jar-resolver` instructions, they say we need to import the files on the export process, quite weird... 

In any case, the current script also generates a unitypackage with EDM4U included.

Subtester/ProjectSettings/ProjectSettings.asset
```diff
@@ -156,7 +156,7 @@ PlayerSettings:
   androidMaxAspectRatio: 2.1
   applicationIdentifier:
     Android: com.revenuecat.purchases_sample
-    iPhone: com.revenuecat.sampleapp
+    iPhone: com.revenuecat.purchases-sample
```
vegaro, line 159: is this an expected change?

Subtester/Packages/manifest.json
vegaro: Regarding updating the version of EDM4U, I am a bit concerned about using that repository... It only has 6 commits and it looks abandoned. It might not be a bad idea to host it ourselves to be honest, but I wonder if we would be breaking the license or something like that. Also, there's no guarantee that's the original plugin code, and not a modified one

Subtester/ProjectSettings/ProjectSettings.asset
aboedo: I think that'd break the iOS app unless we also have one set up in ASC with that bundle id

aboedo: or rather, it'd break the CI that assumes we're using the API key for `sampleapp`

Subtester/Packages/manifest.json
tonidero: Yeah, I'm also a bit concerned about that... I think at this point we could either:

1. Remove the package dependency and just keep EDM4U as a unitypackage and ask devs to import it manually when using purchases-unity as a package. This seems to be the approach Qonversion follows (they ask devs to download it and import it manually to their projects).

2. Host it ourselves. There could be problems with the license as you said and also, there could be a conflict within openupm since I'm not sure how it would handle if I tried to publish a package with the same name as an existing package: https://github.com/openupm/openupm/blob/73af27d67d1553385b2088b293573056f27fa630/data/packages/com.google.external-dependency-manager.yml.  We probably would need to give it a different name.

Honestly, 1 is much simpler for us but would require more work for devs to use our SDK... 2 has more risks and we would also need to maintain this repo. I'm personally leaning towards 1 though... Any thoughts? @vegaro @aboedo

Subtester/ProjectSettings/ProjectSettings.asset
tonidero: I didn't mean to commit that, will revert it. Good catch!

RevenueCat/Editor/RevenueCatPostInstall.cs
aboedo: let's add a note so the next maintainer doesn't wonder the same thing

Subtester/Packages/manifest.json
aboedo: is there any form of post-install trigger we can use to run a script, to do this on behalf of the developer? UPM does have a scripting API

.circleci/config.yml
```diff
@@ -107,14 +107,16 @@ commands:
...
+        type: string
       target:
         type: string
     steps:
       - run:
           name: Build Unity project and export it
           command: |
             /opt/unity/Editor/Unity \
-            -projectPath IntegrationTests \
+            -projectPath << parameters.project >> \
```
aboedo, line 119: this is definitely easier to work with. 

But I do have one concern, based on a lot of painful experience on iOS: 

If we don't have _any_ test projects that use the old integration method, it's almost guaranteed that we're going to break it. 

That's why we ended up with [these multiple projects](https://github.com/RevenueCat/purchases-ios/tree/main/Tests/InstallationTests) on iOS: we kept making updates to something in Cocoapods and Carthage would break, or swift package manager, or whatever. 

So I'd say we need to either: 
- keep one of these using the old `unityPackage`
- have a new one that's a basic copy, using the old `unityPackage`

Either way we just need to know if we break the `unityPackage`

aboedo (commented):
looking awesome!!

.circleci/config.yml
aboedo: maybe that can be a separate PR or something, but I would try to not consider the Epic "done" until we take care of it, because it might be a while before we come back to this and we might have forgotten how it works by then

Subtester/Packages/manifest.json
tonidero: Oh that's an interesting idea! I took a quick look at it and I think it could potentially work but will need to investigate a bit more. I believe we would need to have a script that does something like:
- If EDM4U not found in project (figuring this out can be difficult if the dev has it in a non-conventional location...):
   - Download latest `unitypackage` from EDM4U repo
   - Import `unitypackage` into project (we have to see if we can do this from these scripts)
- If EDM4U found in project, do nothing.

There is some complexity there but if it works, it could make the installation easier using UPM. That said, note that EDM4U only needs to be installed once, but `purchases-unity` can receive upgrades and it shouldn't need updating EDM4U, so I'm wondering if it's worth the extra complexity...

.circleci/config.yml
tonidero: @aboedo We have 2 projects. `Subtester` using purchases-unity through UPM (as a local package) and `IntegrationTests` using a `unitypackage`. Both projects are built and exported to iOS and Android including the APITests so we compile APITests for both projects to ensure the API is the same on both.

That said, `IntegrationTests` needs to have a `unitypackage` first, that's why we use `Subtester` to generate the `unitypackage` first, then import it to `IntegrationTests` and then build and export `IntegrationTests`. 

So basically we don't have a project using a `unitypackage` where we can test manually, only in automated tests. We could have another project where we use the `unitypackage` system manually but I'm worried of overcomplicating the repo and I don't think it's that valuable. Do lmk if you were thinking of something else!

Subtester/Packages/manifest.json
aboedo: well, we can start without it and then if people are having a rough time with EDM4U then figure out workarounds 👍

.circleci/config.yml
aboedo: not having manual tests is fine, I just want to make sure at least one status check fails if we fail to deliver a working `unityPackage`

tonidero:
I just brought back the EDM4U version we were using in Subtester so we are not using the version in openupm anymore. Also had to remove the dependency of the purchases package with EDM4U to avoid issues. So now, if users want to use purchases-unity as a package, they need to also import the latest unitypackage from the EDM4U repo manually. Nothing should change when using our own unitypackage directly.

In the future, we can consider adding our own version of EDM4U within purchases-unity or uploading our own version to openupm if we want to make the process simpler. 

CC @vegaro @aboedo

vegaro (approved):

PR #180, opened 2022-11-14
Fix CircleCI deployment workflow

tonidero:
The deployment workflow stopped working after the changes in #170. A first attempt at fixing it was done in #177 but that caused an issue where the github-release job didn't find the required files in the workspace. Looks like since the branch of the workspace is different (one points to `release-branches` and the other `release-tags`) they are considered different workspaces so it can't share files between them. 

This PR restructures the deployment into 2 workflows. First `deploy-check` which runs deployment tests, initiates the approval job and creates the tag after approval. The second one builds `export-package` again and then performs the `github-release` job. That way, circleci should consider them part of the same workflow and allow them to share files between them.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -306,7 +306,7 @@ workflows:
     jobs:
       - revenuecat/danger
 
-  deploy:
+  deploy-check:
     when:
       not:
         equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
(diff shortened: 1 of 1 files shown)


vegaro (approved):

.circleci/config.yml
```diff
@@ -336,6 +336,17 @@ workflows:
           <<: *release-branches
       - github-release: *release-tags
```
vegaro, line 337: we should remove this

.circleci/config.yml
```diff
@@ -336,6 +336,17 @@ workflows:
...
 
+  deploy:
```
vegaro, line 339: I like that this is split in two deploy workflows now. Maybe we should do the same for other repositories

tonidero: Yeah I think it's better organized this way, specially considering CircleCI doesn't seem to consider jobs after the tag job part of the same workflow from what I've seen.

PR #205, opened 2022-12-27
Update README with UPM instructions

tonidero:
Update README with instructions on how to use the SDK.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,16 +24,18 @@ This version of the SDK doesn't have support for Amazon Store. If you would like
 📮 | Great support - [Help Center](https://revenuecat.zendesk.com) 
 
 ## Getting Started
-There are 2 supported mechanisms to install `purchases-unity`:
-- You can add `purchases-unity` using Unity Package Manager (UPM) through OpenUPM. First, you will need to add [External Dependencies Manager for Unity (EDM4U)](https://github.com/googlesamples/unity-jar-resolver) in your project if you haven't already. To do that:
-   - Download the `external-dependency-manager-latest.unitypackage` from the root of the https://github.com/googlesamples/unity-jar-resolver repo.
-   - Import that `unitypackage` to your project.
+For more detailed information, you can view our complete documentation at [docs.revenuecat.com](https://docs.revenuecat.com/docs/unity). There are 2 supported mechanisms to install `purchases-unity`:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,6 +24,15 @@ This version of the SDK doesn't have support for Amazon Store. If you would like
 📮 | Great support - [Help Center](https://revenuecat.zendesk.com) 
 
 ## Getting Started
+There are 2 supported mechanisms to use `purchases-unity`:
```
tonidero: We don't usually add this in the README of each repo. Instead we update the docs directly. I guess that's the preferred approach right? In any case, I thought I would add this here and can move to the docs later instead after it's been reviewed.

README.md
```diff
@@ -24,6 +24,15 @@ This version of the SDK doesn't have support for Amazon Store. If you would like
...
+- You can add `purchases-unity` using Unity Package Manager (UPM) through OpenUPM. First, you will need to add [External Dependencies Manager for Unity (EDM4U)](https://github.com/googlesamples/unity-jar-resolver) in your project if you haven't already. To do that:
```
aboedo, line 28: you could also just add the URL in regular UPM, right? not sure what the pros and cons are

README.md
aboedo, line 27:
```suggestion
There are 2 supported mechanisms to install `purchases-unity`:
```

README.md
aboedo, line 28: maybe we could have them as `### Unity Package Manager` and `### .unityPackage file` sections for visual clarity?

aboedo (approved):
looks awesome! Left a couple of minor suggestions

README.md
tonidero: Hmm the official EDM4U repo doesn't support UPM though. There is a version in OpenUPM, but that's just someone creating a wrapper around it and it doesn't seem well supported. The "official" way they have is using a `unitypackage`.

README.md
aboedo: mixed feelings, honestly. On one side I feel like it doesn't hurt to overdocument, but in reality it _can_, since it's yet another place to maintain.

aboedo: let's get @codykerns's thoughts on this

tonidero:
Closing this for now in favor of the official documentation, we can revisit adding installation instructions to the README in the future.

PR #209, opened 2023-01-03
Feature parity: subscriber attributes

aboedo:
Added missing subscriber attributes. 
Currently testing on device

### DIFF
--- a/RevenueCat/Scripts/Purchases.cs
+++ b/RevenueCat/Scripts/Purchases.cs
@@ -874,6 +874,42 @@ public partial class Purchases : MonoBehaviour
 
     /**
      * <summary>
+     * Subscriber attribute associated with the CleverTap ID for the user.
+     * Required for the RevenueCat CleverTap integration.
+     * </summary>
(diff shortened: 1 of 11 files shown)


RevenueCat/Scripts/Purchases.cs
```diff
@@ -872,6 +872,42 @@ public void SetAirshipChannelID(string airshipChannelID)
         _wrapper.SetAirshipChannelID(airshipChannelID);
     }
 
+    /**
+     * <summary>
+     * Subscriber attribute associated with the CleverTap ID for the user.
```
tonidero, line 877: NItpick. Since this is a setter, this should be `Sets subscriber attribute...` I think? Same for the others.

tonidero (approved):
Looks good! After it's been tested :shipit:

aboedo:
tested on device for iOS and Android, working on both

RevenueCat/Scripts/Purchases.cs
aboedo: good call, I updated it for all subs attributes

PR #210, opened 2023-01-04
added extra link to the correct file for symlinking sources

aboedo:
Starting the year off with a nice hack 😂 

When using "Symlink local sources" in Subtester, if the SDK is installed through Unity Package Manager, the wrong file is linked. 

This is due to a bug in Unity: https://forum.unity.com/threads/native-sources-inside-package-not-found-after-exporting-project-to-android.1355321/

This PR updates the build.gradle template so that we manually add the right source. The wrong source file will be added too, but that one doesn't really do anything since it points to an invalid path anyway.

### DIFF
--- a/Subtester/Assets/Plugins/Android/mainTemplate.gradle
+++ b/Subtester/Assets/Plugins/Android/mainTemplate.gradle
@@ -75,5 +75,5 @@ android {
 // that does exist here.
 // More info: https://forum.unity.com/threads/native-sources-inside-package-not-found-after-exporting-project-to-android.1355321/
 android.sourceSets.main.java.srcDirs += [
-    '**DIR_UNITYPROJECT**/../RevenueCat/Plugins/Android/PurchasesWrapper.java'
+    '**DIR_UNITYPROJECT**/../RevenueCat/Plugins/Android/'
 ]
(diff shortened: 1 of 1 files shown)


Subtester/Assets/Plugins/Android/mainTemplate.gradle
```diff
@@ -64,4 +64,16 @@ android {
...
 **EXTERNAL_SOURCES**
+
+// IMPORTANT: If you're not installing through Unity Package Manager, remove the code below.
+// Symlink local sources. When using Unity's "Symlink local sources" checkbox,
+// The link will be wrong when using Unity Package Manager because of a bug in Unity.
+// That's not a problem if it points to a file that doesn't exist, but we need to add the path
+// that does exist here.
+// More info: https://forum.unity.com/threads/native-sources-inside-package-not-found-after-exporting-project-to-android.1355321/
+android.sourceSets.main.java.srcDirs += [
+    '**DIR_UNITYPROJECT**/../RevenueCat/Plugins/Android/PurchasesWrapper.java'
```
vegaro, line 78: Nice hack I like it

tonidero: Ahh nice! I guess we can leave the wrong path yeah 😅.

Also, we could make it `'**DIR_UNITYPROJECT**/../RevenueCat/Plugins/Android/*.java'` to include other files if we ever add any, but considering this is a hack until Unity fixes it (if ever) then I think this is ok.

aboedo: that's a good idea, I'll do that

aboedo: doing `*.java` wouldn't work, but I just made it `.....RevenueCat/Plugins/Android/` instead and it worked fine, going with that

PR #222, opened 2023-01-20
Add StoreProduct.SubscriptionPeriod

vegaro:
[CSDK-593](https://revenuecats.atlassian.net/browse/CSDK-593)

[CSDK-593]: https://revenuecats.atlassian.net/browse/CSDK-593?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Subtester/Assets/Scripts/PurchasesListener.cs
+++ b/Subtester/Assets/Scripts/PurchasesListener.cs
@@ -70,7 +70,7 @@ public class PurchasesListener : Purchases.UpdatedCustomerInfoListener
                 {
                     Debug.Log("Package " + package);
                     if (package == null) continue;
-                    var label = package.PackageType + " " + package.StoreProduct.PriceString;
+                    var label = package.PackageType + " " + package.StoreProduct.PriceString + "\n" + package.StoreProduct.SubscriptionPeriod;
                     CreateButton(label, () => PurchasePackageButtonClicked(package));
                 }
             }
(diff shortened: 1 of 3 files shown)


Subtester/Assets/Scripts/PurchasesListener.cs
```diff
@@ -70,7 +70,7 @@ private void CreatePurchasePackageButtons()
                 {
                     Debug.Log("Package " + package);
                     if (package == null) continue;
-                    var label = package.PackageType + " " + package.StoreProduct.PriceString;
+                    var label = package.PackageType + " " + package.StoreProduct.PriceString + "\n" + package.StoreProduct.SubscriptionPeriod;
```
tonidero, line 73: Hmm I guess it will be pretty self-explanatory but I wonder if we should add a label to the period, like `SubscriptionPeriod: P1Y`. Mostly for people not as familiar with this. This wouldn't apply to inapp's though, where adding that might make it more confusing... I'm ok leaving as is though

tonidero (approved):

Subtester/Assets/Scripts/PurchasesListener.cs
vegaro: I actually added this for test purposes only, and then I noticed there is a log at the top of the app so I might just remove this completely

PR #231, opened 2023-01-31
Update gems

vegaro:
We got some dependabot alerts of high severity vulnerabilities https://github.com/RevenueCat/purchases-unity/security/dependabot


tonidero (approved):
Oh I don't have access to that it seems, but this makes sense 👍

PR #237, opened 2023-02-09
Adds Purchases.SetLogHandler

vegaro:
[CSDK-558]

[CSDK-558]: https://revenuecats.atlassian.net/browse/CSDK-558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RevenueCat/Plugins/Android/com/revenuecat/purchasesunity/purchasesWrapper.kt
```diff
@@ -0,0 +1,12 @@
...
+import com.revenuecat.purchases.hybridcommon.mappers.convertToJson
+import com.revenuecat.purchases.hybridcommon.setLogHandler
+
+private const val HANDLE_LOG = "_handleLog"
+
+fun setLogHandler() {
+    setLogHandler {
+        PurchasesWrapper.sendJSONObject(it.convertToJson(), HANDLE_LOG)
+    }
+}
```
vegaro, line 12: This is an alternative to use the new function added in https://github.com/RevenueCat/purchases-hybrid-common/pull/328

Adding Kotlin works fine, the only problem I have seen is that after exporting the project, it adds Kotlin 1.4.31, and trying to compile the project gives an error that requires updating to 1.6.0. I think it's due to incompatibilities either with the Gradle version or the Android Gradle Plugin.

Do you think it's a deal breaker?

vegaro: cc @RevenueCat/coresdk

RevenueCat/Plugins/Android/com/revenuecat/purchasesunity.meta
```diff
@@ -0,0 +1,8 @@
+fileFormatVersion: 2
```
vegaro, line 1: I moved the files to the correct folder `com/revenuecat/purchasesunity` so their package matches the folder

RevenueCat/Plugins/Android/com/revenuecat/purchasesunity/purchasesWrapper.kt
tonidero: Hmm would this affect developers in their own projects? If so, I think it would be.

This incompatibility is between the unity sdk and the `purchases-android` SDK? If so, I guess that means our android SDK is not compatible with that old version of kotlin...

vegaro: I think it's between purchases-android or purchases-hybrid-common, I am not sure, but yes, I think it's in general an incompatibility between the version of the kotlin build plugin and the version we used to compile

vegaro: And yes, this would mean people exporting their unity project with this version of Unity, which enables Kotlin 1.4.31 need to update their build.gradle

tonidero: We could also try to make our android sdk compatible with that version of kotlin... but not sure if that's possible

vegaro: By the way, I tested Android's MagicWeather and looks like we require Kotlin 1.5.0+, using a kotlin version lower than that gives the same error

RevenueCat/Plugins/Android/PurchasesWrapper.java
```diff
@@ -386,8 +403,7 @@ private static void logJSONException(JSONException e) {
         Log.e("Purchases", "JSON Error: " + e.getLocalizedMessage());
     }
 
-    private static void sendJSONObject(JSONObject object, String method) {
-        Log.e("Purchases", object.toString());
```
vegaro, line 390: removed these logs because they were making the logcat super poluted

RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
```diff
@@ -365,14 +372,8 @@ - (void)sendJSONObject:(NSDictionary *)jsonObject toMethod:(NSString *)methodNam
...
     NSData *responseJSONData = [NSJSONSerialization dataWithJSONObject:jsonObject options:0 error:&error];
 
-    if (error) {
-        NSLog(@"Error serializing products: %@", error.localizedDescription);
-        return;
-    }
-
     if (responseJSONData) {
         NSString *json = [[NSString alloc] initWithData:responseJSONData encoding:NSUTF8StringEncoding];
-        NSLog(@"json = %@", json);
```
vegaro, line 375: Also removed these logs because they were poluting the logs a lot

RevenueCat/Plugins/Android/PurchasesWrapper.java
tonidero: Right, additionally, we should be careful with all these hybrid side logs. As I mentioned before, ideally they would be rerouted to the same log handler as well so devs can handle all logs.

I think this can be done in a different PR but we definitely shouldn't lose track of it.

tonidero (approved):
LGTM! Just that concern about hybrid side logs that we can address later.

PR #245, opened 2023-02-21
Fix for `NSInvalidArgumentException` in `GetPromotionalOffer`

vegaro:
We were not handling errors in `GetPromotionalOffer` and we were trying to send a nil response to the Unity message layer

[TRIAGE-276]

[TRIAGE-276]: https://revenuecats.atlassian.net/browse/TRIAGE-276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
+++ b/RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
@@ -265,14 +265,11 @@ signedDiscountTimestamp:(NSString *)signedDiscountTimestamp {
     [RCCommonFunctionality promotionalOfferForProductIdentifier:productIdentifier
                                                       discount:discountIdentifier
                                                completionBlock:^(NSDictionary *_Nullable responseDictionary, RCErrorContainer *_Nullable error) {
-        NSMutableDictionary *response;
-        if (error) {
-            response = [NSMutableDictionary new];
-            response[@"error"] = error.info;
-
-        } else {
-            response = [NSMutableDictionary dictionaryWithDictionary:responseDictionary];
+        NSDictionary *response = (error)
+        ? @{
+            @"error": error.info
         }
+        : responseDictionary;
         [self sendJSONObject:response toMethod:GET_PROMOTIONAL_OFFER];
     }];
 }
--- a/RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
+++ b/RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
@@ -265,7 +265,15 @@ signedDiscountTimestamp:(NSString *)signedDiscountTimestamp {
     [RCCommonFunctionality promotionalOfferForProductIdentifier:productIdentifier
                                                       discount:discountIdentifier
                                                completionBlock:^(NSDictionary *_Nullable responseDictionary, RCErrorContainer *_Nullable error) {
-        [self sendJSONObject:responseDictionary toMethod:GET_PROMOTIONAL_OFFER];
+        NSMutableDictionary *response;
+        if (error) {
+            response = [NSMutableDictionary new];
+            response[@"error"] = error.info;
+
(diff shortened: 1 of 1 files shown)


RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
```diff
@@ -265,7 +265,15 @@ - (void)promotionalOfferForProductIdentifier:(NSString *)productIdentifier
...
                                                       discount:discountIdentifier
                                                completionBlock:^(NSDictionary *_Nullable responseDictionary, RCErrorContainer *_Nullable error) {
-        [self sendJSONObject:responseDictionary toMethod:GET_PROMOTIONAL_OFFER];
+        NSMutableDictionary *response;
+        if (error) {
+            response = [NSMutableDictionary new];
+            response[@"error"] = error.info;
+
+        } else {
+            response = [NSMutableDictionary dictionaryWithDictionary:responseDictionary];
```
joshdholtz, line 274: Do we need to make sure that `responseDictionary` isn't `nil` at all? Or is it safe to assume it never will be?

joshdholtz (approved):
One small question but looks good!

RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
NachoSoto: Yeah I'd create the `NSMutableDictionary` in line 268 so we don't risk sending `nil`.

RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
```diff
@@ -265,7 +265,15 @@ - (void)promotionalOfferForProductIdentifier:(NSString *)productIdentifier
...
+        }
+        [self sendJSONObject:response toMethod:GET_PROMOTIONAL_OFFER];
```
NachoSoto, line 276: Maybe send an immutable copy?
```suggestion
        [self sendJSONObject:response.copy toMethod:GET_PROMOTIONAL_OFFER];
```

NachoSoto: Actually I think you could simplify this while thing like this:
```objc
NSDictionary *response = (error)
? @{
  @"error": error.info
}
: responseDictionary;
```

joshdholtz: I don't understand this language

RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
vegaro: Initializing to an empty dictionary will break the unity code where we try to create the classes from the json response. I think we should create an error here, but the error should be in the purchases hybrid common side so all hybrids are protected from that unexpected case where both the response and the error are nil

vegaro: Oh actually, for incompatible versions phc sends nil for both 🤔

vegaro: I opened this PR https://github.com/RevenueCat/purchases-hybrid-common/pull/338/ to stop sending two nils, and also created an error in this PR if both responses and error are nil

joshdholtz (approved):
LGTM 💪

RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
tonidero: Hmm right, that seems to be the same it is right now (with different indentation)?

tonidero (approved):

PR #249, opened 2023-02-23
Runs integration tests on main too

vegaro:
This https://github.com/RevenueCat/purchases-unity/pull/237 broke integration tests

Running these on every PR would be expensive and would slow down merging too much, since they are slow, but running them on main would help determine which PR  broke them if they break


tonidero (approved):
Well, this would save us from running it on every commit, only when merging PRs. I think it's a good middle-ground 👍

PR #293, opened 2023-06-22
BC5 Support

MarkVillacampa:
## Motivation

Support Purchases Android V6 which includes support for BC5.

## Tasks

- [x] Update purchases-hybrid-common dependency to the latest version.
- [x] Update PurchasesWrapper.java to support purchasing non-backwards-compatible products.
- [x] Update Purchases.cs/PurchasesWrapper.cs/PurchasesWrapperAndroid.cs to support purchasing non-backwards-compatible products.
- [x] Add supporting C# classes and deserialize PHC JSON (SubscriptionOption, PricingPhase, etc).
- [x] Add support for Offerings Metadata.
- [x] Add news APIs to IntegrationTests
- [X] Map SubscriptionOption.PricingPhases (#298) via Josh Holtz (@joshdholtz)
- [X] Add buttons to purchase all SubscriptionOptions in the Subtester app
- [X] Add a way to test Product Changes in the Subtester app


.circleci/config.yml
```diff
@@ -385,19 +385,16 @@ workflows:
...
       - build-integration-tests-android:
-          <<: *release-branches-and-main
           requires:
             - export-package
       - build-integration-tests-ios:
-          <<: *release-branches-and-main
           requires:
             - export-package
       - archive-ios:
-          <<: *release-branches
```
MarkVillacampa, line 400: Enabling builds on PRs. Takes <20 min to run but there is room for improvement here.

RevenueCat/Scripts/Utilities.cs
```diff
@@ -23,7 +23,7 @@ internal static DateTime FromUnixTimeInMilliseconds(long unixTimeInMilliseconds)
 
         internal static String DictToString<T>(Dictionary<string, T> dictionary)
         {
-            var items = dictionary.Select(kvp => string.Format($"{kvp.Key} : Id={kvp.Value.ToString()}"));
+            var items = dictionary.Select(kvp => $"{kvp.Key} : Id={kvp.Value.ToString()}");
```
MarkVillacampa, line 26: `string.Format` was causing an exception when printing an object with two nested dictionaries where the inner one was empty. Since we're not actually using the format functionality, instead we're interpolating the strings, we can safely remove it.

The code was effectively doing:

```
string.Format(string.Format("{{}}"));
```

Which throws:

```
FormatException: Input string was not in a correct format.
                                                                                                      at System.Text.StringBuilder.FormatError () [0x00000] in <00000000000000000000000000000000>:0 
                                                                                                      at System.Text.StringBuilder.AppendFormatHelper (System.IFormatProvider provider, System.String format, System.ParamsArray args) [0x00000] in <00000000000000000000000000000000>:0 
                                                                                                      at System.String.FormatHelper (System.IFormatProvider provider, System.String format, System.ParamsArray args) [0x00000] in <00000000000000000000000000000000>:0 
                                                                                                      at System.String.Format (System.String format, System.Object[] args) [0x00000] in <00000000000000000000000000000000>:0 
                                                                                                      at PurchasesListener.GetOfferings () [0x00000] in <00000000000000000000000000000000>:0 
                                                                                                      at UnityEngine.Events.UnityEvent.Invoke () [0x00000] in <00000000000000000000000000000000>:0 
                                                                                                      at UnityEngine.EventSystems.ExecuteEvents.Execute[T] (UnityEngine.GameObject target, UnityEngine.EventSystems.BaseEventData eventData, UnityEngine.EventSystems.ExecuteEvents+EventFunction`1[T1] functor) [0x00000
```

RevenueCat/Scripts/SubscriptionOption.cs
```diff
@@ -0,0 +1,307 @@
...
+                    $"{nameof(RecurrenceMode)}: {RecurrenceMode}\n" +
+                    $"{nameof(BillingCycleCount)}: {BillingCycleCount}\n" +
+                    $"{nameof(Price)}: {Price}\n" +
+                    $"{nameof(OfferPaymentMode)}: {OfferPaymentMode}\n";
+            }
+        }
+
+        public class Period
+        {
+            public readonly PeriodUnit Unit;
```
vegaro, line 185: This is missing docs

```    
/**
* The number of period units: day, week, month, year, unknown
*/
```

RevenueCat/Scripts/SubscriptionOption.cs
```diff
@@ -0,0 +1,307 @@
...
+    /// Contains all details associated with a SubscriptionOption
+    /// Used only for Google
+    public class SubscriptionOption
+    {
+        /**
+         * Identifier of the subscription option
+         * If this SubscriptionOption represents a base plan, this will be the basePlanId.
+         * If it represents an offer, it will be {basePlanId}:{offerId}
+         */
+        public readonly string ID;
```
tonidero, line 17: Should this be `Id`?

RevenueCat/Scripts/ProductCategory.cs
```diff
@@ -0,0 +1,14 @@
+public partial class Purchases
+{
+    public enum ProductCategory
```
tonidero, line 3: Same for this, we will need new methods in `Purchases.cs` to use these instead of the current versions?

MarkVillacampa: This is a new property in `StoreProduct` returned by the API, we don't need to pass it to any of the methods in Purchases.cs

RevenueCat/Scripts/GoogleProductChangeInfo.cs
```diff
@@ -0,0 +1,19 @@
+using JetBrains.Annotations;
+
+public partial class Purchases
+{
+    /// <summary>
+    /// Holds the information used when upgrading from another sku. For Android use only.
+    /// </summary>
+    public class GoogleProductChangeInfo
```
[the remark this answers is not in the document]
MarkVillacampa: 👍 Let's keep the methods as they are for now (it's only a couple parameters really) and tackle the builder pattern in all hybrids at the same time.

RevenueCat/Scripts/Offering.cs
```diff
@@ -13,14 +14,15 @@ public class Offering
         public readonly string Identifier;
         public readonly string ServerDescription;
         public readonly List<Package> AvailablePackages;
+        [CanBeNull] public readonly Dictionary<string, object> Metadata;
```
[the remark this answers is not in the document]
MarkVillacampa: I'll add one for strings at least! we now support deeply nested and arbitrary types in metadata, so adding helpers for each type can become out of hand quickly 😅

RevenueCat/Scripts/ProductCategory.cs
tonidero: Ohh you're right... I thought we were using `ProductType`, but seems like we aren't in unity, so we don't need to deprecate it.

RevenueCat/Plugins/Android/PurchasesWrapper.java
```diff
@@ -128,19 +133,23 @@ public void onError(ErrorContainer errorContainer) {
...
+        purchaseProduct(productIdentifier, type, null,  0, false, null);
     }
 
     public static void purchasePackage(String packageIdentifier,
                                        String offeringIdentifier,
                                        @Nullable final String oldSKU,
-                                       final int prorationMode) {
+                                       final int prorationMode,
+                                       final boolean isPersonalized,
+                                       @Nullable final String presentedOfferingIdentifier
```
tonidero, line 144: This parameter seems duplicated and unused, there is already an `offeringIdentifier` above, which should be the same I think?

RevenueCat/Scripts/PurchasesWrapper.cs
```diff
@@ -10,11 +10,15 @@ public interface IPurchasesWrapper
 
     void PurchaseProduct(string productIdentifier, string type = "subs", string oldSku = null,
         Purchases.ProrationMode prorationMode = Purchases.ProrationMode.UnknownSubscriptionUpgradeDowngradePolicy,
+        bool googleIsPersonalizedPrice = false, string presentedOfferingIdentifier = null,
```
tonidero, line 13: These parameters seem to not be used by callers currently? I guess we are planning to offer this in the API in the future, but currently is unused correct? Maybe add a comment mentioning that until that part is done? Same in the `PurchasePackage` function

.circleci/config.yml
tonidero: Hmm I think it's ok. We can keep an eye out to make sure costs don't grow too much.

tonidero (approved):
Left a couple comments but nothing blocking I think.

RevenueCat/Plugins/Android/PurchasesWrapper.java
MarkVillacampa: Indeed this was duplicated! Since it is used internally only I can remove it without changing the public api

RevenueCat/Scripts/PurchasesWrapper.cs
MarkVillacampa: Forgot to expose `googleIsPersonalizedPrice` in `Purchases.cs` will add it there.

PR #322, opened 2023-09-13
Fix for iOS discounts not being returned

joshdholtz:
## Motivation

[DENG-534](https://linear.app/revenuecat/issue/DENG-534/unity-sdk-ios-discounts-arent-returned)

iOS discounts were not being returned

## Description

Logic for Google Play subscription options had an early return if they weren't found which lead to the skipping of iOS discount logic

- Removed an early return in subscription option logic
  - Replaced with an else to set the subscription options
- Removed an early return in discount logic
  - Replaced with an else to set the discounts

### DIFF
--- a/RevenueCat/Scripts/StoreProduct.cs
+++ b/RevenueCat/Scripts/StoreProduct.cs
@@ -62,31 +62,36 @@ public partial class Purchases
             {
                 DefaultOption = new SubscriptionOption(defaultOptionJsonNode);
             }
+
             var subscriptionOptionsResponse = response["subscriptionOptions"];
-            if (subscriptionOptionsResponse == null)
+            if (subscriptionOptionsResponse == null) 
             {
                 SubscriptionOptions = null;
-                return;
             }
-            var subscriptionOptionsTemporaryList = new List<SubscriptionOption>();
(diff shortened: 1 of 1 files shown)


RevenueCat/Scripts/StoreProduct.cs
```diff
@@ -62,31 +62,36 @@ public StoreProduct(JSONNode response)
             {
                 DefaultOption = new SubscriptionOption(defaultOptionJsonNode);
             }
+
             var subscriptionOptionsResponse = response["subscriptionOptions"];
-            if (subscriptionOptionsResponse == null)
+            if (subscriptionOptionsResponse == null)
```
aboedo, line 67: we can kill this trailing space. Doesn't matter though

aboedo (approved):
thanks for taking care of this so quickly!

RevenueCat/Scripts/StoreProduct.cs
joshdholtz: That was my favorite space

MarkVillacampa (approved):

tonidero (approved):

PR #338, opened 2023-10-04
Add support for configuring in-app messages

MarkVillacampa:


RevenueCat/Scripts/Purchases.cs
```diff
@@ -98,6 +98,12 @@ public partial class Purchases : MonoBehaviour
               "you which StoreKit implementation to use.", false)]
     public bool usesStoreKit2IfAvailable;
 
+    [Tooltip("Whether we should show store in-app messages automatically. Both Google Play and the App Store provide in-app " +
+             "messages for some situations like billing issues. By default, those messages will be shown automatically.\n" +
+             "This allows to disable that behavior, so you can display those messages at your convenience. For more information, " +
+             "check: https://rev.cat/storekit-message and https://rev.cat/googleplayinappmessaging")]
+    public bool shouldShowInAppMessagesAutomatically;
```
tonidero, line 105: Can we default this to `true`?

tonidero (approved):
Just one thing, to make sure the defaults are correct. Other than that, looks good!

PR #345, opened 2023-10-19
Add BillingClient versions to VERSIONS table

vegaro:
Added BillingClient version of the Unity IAP plugin since it simplified some research I was doing.

Also added version 5.3.1 which was missing.


tonidero (approved):
❤️

PR #370, opened 2023-11-20
`Trusted entitlements`: Add support for trusted entitlements

tonidero:
See http://rev.cat/trusted-entitlements for more info on this feature.

### New APIs:
-  `EntitlementVerificationMode` and `VerificationResult` enums
- `PurchasesConfiguration` new setter method `SetEntitlementVerificationMode`
- `EntitlementInfos` and `EntitlementInfo` new property `Verification`

### DIFF
--- a/RevenueCat/Scripts/EntitlementVerificationMode.cs
+++ b/RevenueCat/Scripts/EntitlementVerificationMode.cs
@@ -1,0 +1,45 @@
+using System;
+using System.ComponentModel;
+
+public partial class Purchases
+{
+    /// <summary>
+    /// Enum of entitlement verification modes.
+    /// </summary>
+    public enum EntitlementVerificationMode
+    {
+        /// <summary>
+        /// The SDK will not perform any entitlement verification.
+        /// </summary>
+        [Description("DISABLED")]
+        Disabled,
+
+        /// <summary>
+        /// Enable entitlement verification.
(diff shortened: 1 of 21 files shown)


RevenueCat/Scripts/EntitlementVerificationMode.cs
```diff
@@ -0,0 +1,45 @@
...
+
+internal static class EntitlementVerificationModeExtensions
+{
+    internal static string Name(this Purchases.EntitlementVerificationMode verificationMode)
+    {
+        var type = verificationMode.GetType();
+        var memInfo = type.GetMember(verificationMode.ToString());
+        var attributes = memInfo[0].GetCustomAttributes(typeof(DescriptionAttribute), false);
+        var stringValue = ((DescriptionAttribute) attributes[0]).Description;
+        return stringValue;
```
vegaro, line 43: If I understand correctly, this is accessing the Description annotation to return the value correct?

vegaro (approved):

RevenueCat/Scripts/EntitlementVerificationMode.cs
tonidero: Yes, we used this for mapping the `LogLevel` enum, and I thought it was better than mapping integers, even if very convoluted 😅

PR #376, opened 2023-11-28
Fix `isAnonymous` in iOS

vegaro:
We were not returning anything in iOS for `isAnonymous`, which was evaluating to `true`.

https://linear.app/revenuecat/issue/SDK-3285/fix-isanonymous-in-ios-unity


tonidero (approved):
Nice catch! ❤️

PR #471, opened 2024-06-28
Add visionOS support

aboedo:
Draft adding visionOS support

### DIFF
--- a/RevenueCat/Editor/RevenueCatPostInstall.cs
+++ b/RevenueCat/Editor/RevenueCatPostInstall.cs
@@ -1,5 +1,5 @@
 
-#if UNITY_EDITOR && UNITY_IOS
+#if UNITY_EDITOR && (UNITY_IOS || UNITY_VISIONOS)
 
 using System.IO;
 using UnityEngine;
--- a/RevenueCat/Scripts/Purchases.cs
+++ b/RevenueCat/Scripts/Purchases.cs
@@ -113,7 +113,7 @@ public partial class Purchases : MonoBehaviour
     {
 #if UNITY_ANDROID && !UNITY_EDITOR
         _wrapper = new PurchasesWrapperAndroid();
-#elif UNITY_IPHONE && !UNITY_EDITOR
+#elif (UNITY_IOS || UNITY_VISIONOS) && !UNITY_EDITOR
         _wrapper = new PurchasesWrapperiOS();
 #else
         _wrapper = new PurchasesWrapperNoop();
--- a/RevenueCat/Scripts/PurchasesWrapperiOS.cs
+++ b/RevenueCat/Scripts/PurchasesWrapperiOS.cs
@@ -3,7 +3,7 @@ using System.Runtime.InteropServices;
 using JetBrains.Annotations;
 using UnityEngine;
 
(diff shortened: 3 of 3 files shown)


RevenueCat/Scripts/Purchases.cs
```diff
@@ -113,7 +113,7 @@ private void Start()
     {
 #if UNITY_ANDROID && !UNITY_EDITOR
         _wrapper = new PurchasesWrapperAndroid();
-#elif UNITY_IPHONE && !UNITY_EDITOR
+#elif (UNITY_IOS || UNITY_VISIONOS) && !UNITY_EDITOR
```
aboedo, line 116: this would be the reason why before this PR, it would still compile, but no-op

RevenueCat/Editor/RevenueCatPostInstall.cs
```diff
@@ -1,5 +1,5 @@
 
-#if UNITY_EDITOR && UNITY_IOS
+#if UNITY_EDITOR && (UNITY_IOS || UNITY_VISIONOS)
```
tonidero, line 2: I haven't been able to test this, but do we need to also check for vision os in line 16 below? https://docs.unity3d.com/ScriptReference/BuildTarget.VisionOS.html

tonidero (approved):
Just a comment but LGTM!

RevenueCat/Editor/RevenueCatPostInstall.cs
AstoraGray: Does it support Unity-VisionOS now?

tonidero: After testing with the new `BuildTarget.VisionOS`, this actually brakes with the current unity version we are using and it would be considered a breaking change. In any case, this script shouldn't really be needed in modern unity versions so we can merge without it.

AstoraGray:
Hello, our team encountered the following error while packaging RevenueCat for Unity version 7.2.0 to Vision Pro using Xcode. The error did not occur when rolling back to version 7.1.2, and the packaging was successful, so there should still be some issues with version 7.2.0's support for Vision Pro.
image

aboedo:
@AstoraGray thanks for letting us know! I'll reopen this ticket and we'll take care of it

aboedo:
Update: turns out this is a bit trickier than expected since EDM4U, which we use to manage iOS dependencies in Unity, [doesn't support VisionOS yet](https://github.com/googlesamples/unity-jar-resolver/issues/628). 

We're studying alternatives

AstoraGray:
> Update: turns out this is a bit trickier than expected since EDM4U, which we use to manage iOS dependencies in Unity, [doesn't support VisionOS yet](https://github.com/googlesamples/unity-jar-resolver/issues/628).
> 
> We're studying alternatives

I wanted to discuss the recent updates regarding the visionOS support that was merged into the main branch recently. As our project team continues to navigate the integration of this feature, I wanted to kindly inquire if it would be feasible to complete the development by the end of this year.

This functionality is crucial for our project, and having it finalized by then would greatly assist us in meeting our upcoming deadlines.

PR #476, opened 2024-07-23
Update `VERSIONS.md` to include Billing client version and update fastlane plugin

tonidero:
We updated the fastlane plugin to add the billing client version to the VERSIONS.md file. This should be done automatically in the other hybrids but in Unity, we need to do this manually since there is an extra column in this file added here.

This performs this update so it works moving forward.


VERSIONS.md
```diff
@@ -1,99 +1,100 @@
...
-| 2.0.6         | 3.0.1       | 3.0.4           | 1.0.6          | Unity IAP < 3.1.0             |
-| 2.0.5         | 3.0.1       | 3.0.3           | 1.0.6          | Unity IAP < 3.1.0             |
-| 2.0.4         | 3.0.0       | 3.0.3           | 1.0.1          | Unity IAP < 3.1.0             |
-| 2.0.3         | 3.0.0       | 3.0.2           | 1.0.1          | Unity IAP < 3.1.0             |
-| 2.0.0         | 3.0.0       | 3.0.2           | 1.0.1          | Unity IAP < 3.1.0             |
-| 1.2.2         | 2.5.0       | 2.4.1           |                |                               |
-| 1.2.1         | 2.5.0       | 2.3.0           |                |                               |
+| Version       | iOS version | Android version | Common version | Play Billing Library version | Unity IAP compatibility       |
+| ------------- | ----------- | --------------- | -------------- | ---------------------------- | ----------------------------- |
+| 6.10.0        | 4.43.2      | 7.12.0          | 11.1.0         | 6.2.1                        | Unity IAP 4.12.0+ (BC6)       |
```
tonidero, line 3: I thought about also adding links to releases for all these releases (as added in the plugin)... But I think it's ok to just add them moving forward.

vegaro (approved):

JayShortway (approved):

PR #479, opened 2024-07-25
Fix `Gemfile.lock` with new fastlane plugin dependencies

tonidero:
I didn't update the fastlane plugin correctly in #476 so it was causing some failures. This fixes that by adding new dependencies of the fastlane plugin.


JayShortway (approved):
LGTM!

PR #482, opened 2024-07-30
API improvements

Legoless:
- [x] A description about what and why you are contributing, even if it's trivial.

This PR addresses two API issues:
1. Purchases Listener is currently an abstract `MonoBehaviour`, which ties the behaviour to Unity components. This is not bad, but it does limit you, which is entirely not needed, if you mostly use the API from code. There's no other requirement that this should in fact be an abstract class. So a new interface is introduced instead and backwards support is added, so that the `UpdatedCustomerInfoListener` implements the interface.

2. The whole `Purchases` object which is required for RevenueCat to operate initializes itself fairly late in `Start` function. This is a problem, if you need to do anything beforehand as we're booting our game up in `Awake` method. The script order is usually undefined and we've seen many differences in it between real devices and Editor, so sometimes `Purchases` would prepare itself for API type of configuration, sometimes we would be too early calling `Configure` and crash with `NullPointerException` on internal `_wrapper`. To get around this problem, we extracted initialization of wrapper to a separate function that only creates the instances once, but can be called many times. That way `Configure` will `Prepare` the wrapper, before it calls to it's `Setup` method.

- [x] The issue number(s) or PR number(s) in the description if you are contributing in response to those.
 No open issues right now for these improvements.

- [x] If applicable, unit tests.
No additional Unit Tests required.

### DIFF
--- a/RevenueCat/Scripts/Purchases.cs
+++ b/RevenueCat/Scripts/Purchases.cs
@@ -58,7 +58,7 @@ public partial class Purchases : MonoBehaviour
 
     [Tooltip("A subclass of Purchases.UpdatedCustomerInfoListener component.\n" +
              "Use your custom subclass to define how to handle updated customer information.")]
-    public UpdatedCustomerInfoListener listener;
+    public IUpdatedCustomerInfoListener listener;
 
     [Tooltip("An optional boolean. Set this to true if you have your own IAP implementation " +
              "and want to use only RevenueCat's backend.\nDefault is false.\n" +
@@ -111,22 +111,32 @@ public partial class Purchases : MonoBehaviour
 
     private void Start()
     {
-#if UNITY_ANDROID && !UNITY_EDITOR
-        _wrapper = new PurchasesWrapperAndroid();
-#elif UNITY_IPHONE && !UNITY_EDITOR
-        _wrapper = new PurchasesWrapperiOS();
-#else
-        _wrapper = new PurchasesWrapperNoop();
-#endif
-        if (!string.IsNullOrEmpty(proxyURL))
-        {
-            _wrapper.SetProxyURL(proxyURL);
-        }
+        Prepare();
 
         if (useRuntimeSetup) return;
 
         Configure(string.IsNullOrEmpty(appUserID) ? null : appUserID);
         GetProducts(productIdentifiers, null);
+    }
+    
+    private void Prepare()
+    {
+        if (_wrapper != null)
+        {
+            return;
+        }
+        #if UNITY_ANDROID && !UNITY_EDITOR
+        _wrapper = new PurchasesWrapperAndroid();
+        #elif UNITY_IPHONE && !UNITY_EDITOR
+        _wrapper = new PurchasesWrapperiOS();
+        #else
+        _wrapper = new PurchasesWrapperNoop();
+        #endif
+        
+        if (!string.IsNullOrEmpty(proxyURL))
+        {
+            _wrapper.SetProxyURL(proxyURL);
+        }
     }
 
     private void Configure(string newUserId)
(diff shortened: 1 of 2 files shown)


RevenueCat/Scripts/Purchases.cs
```diff
@@ -58,7 +58,26 @@ public partial class Purchases : MonoBehaviour
 
     [Tooltip("A subclass of Purchases.UpdatedCustomerInfoListener component.\n" +
              "Use your custom subclass to define how to handle updated customer information.")]
-    public UpdatedCustomerInfoListener listener;
+    [SerializeField]
+    private UpdatedCustomerInfoListener listener;
+    
+    // Cache custom listener, if not from class.
+    private IUpdatedCustomerInfoListener customListener;
+    public IUpdatedCustomerInfoListener Listener
```
tonidero, line 66: I believe this means that devs setting the listener through code before might experience a breaking change.
- Before:
```
purchases.listener = CustomListener()
```

- After:
```
purchases.Listener = CustomListener()
```

I do think that most people would set the listener through the editor, considering it had to be a `MonoBehavior`... But I think we might need to make the `listener` public private still to avoid the breaking change, and encourage people to move to use `Listener` instead. Wdyt?

tonidero (commented):
Hi @Legoless,

Really sorry for the delay, and thanks for your contribution! The changes here make a lot of sense. I just left a comment about a possible breaking change. What do you think?

Once we get that sorted I can get this merged

RevenueCat/Scripts/Purchases.cs
tonidero: Maybe we can hold on this change until the next major as well, and release the change below

RevenueCat/Scripts/Purchases.cs
```diff
@@ -111,22 +130,32 @@ public partial class Purchases : MonoBehaviour
 
     private void Start()
     {
-#if UNITY_ANDROID && !UNITY_EDITOR
+        Prepare();
```
tonidero, line 133: This is great! Makes a lot of sense, thanks for adding this!

RevenueCat/Scripts/Purchases.cs
Legoless: You're correct, this is breaking change in code. I decided for this approach for two reasons:
- It's very easy to search and replace in code: `Listener` instead of `listener`. Could leave some comments and hint maybe for those people that use it in code.
- It's very clear from API design perspective that the private field is meant for Editor only and you only allow setting it there. In other way you would have two exposed `listener` and `Listener` in the API, which could get confusing, if in fact you did want to set it in code as a new user.

My suggestion would be to keep the API clearer and rather make a major release (or however you think is best), if this breaking change is a problem, since this move will make the API clearer in the future. But at the end it's your call how you want to have it, definitely both solutions work for our case. Maybe a better approach would be to have a `List` of those listeners instead and keep the original one public and just add like a `AddListener`/`RemoveListener` methods to manage the list.. Just thinking out loud.

Your call, let me know if you need me to change in the PR.

tonidero: Yeah I agree that it would be better to ship this in a major, rather than keeping both which would cause a very confusing API.

If you want to revert this change and leave the one below, we would be happy to get that in!

As for your proposal of having `AddListener/`RemoveListener`, I think that could work! We actually allow doing this in some other SDKs. But seems that could be worked on separately. Thanks for the suggestion!

Legoless: I can also do the `Add/Remove`Listener if you like, but not sure which change you want me to revert? The change to `private`?

tonidero: Sorry for not being clear! My suggestion is to:
- Revert the changes related to the listener (including the new interface)
- Keep the changes related to initialization of the SDK at different points of time.

Then merge the PR with just that second change, which shouldn't be a breaking change. And we can release that soon.

As for doing the `Add/Remove` listener, I think that works too! And it would be amazing if you can do it 👼. The only point is to avoid breaking changes (at least until we're ready to make another major)

Please let me know if you have questions or help with the listener approach! We also think that would be a good addition 👍

Legoless: Cool, I understand, I'll get on it this week.

StephenHodgson:
I've completed a major refactor to address a lot of the problems pointed out in this pull request as well as in #612 in #674

PR #485, opened 2024-08-01
Add support new native majors

tonidero:
This bumps PHC to 13.0.0 which includes breaking changes in the native SDKs: 
- iOS v5: https://github.com/RevenueCat/purchases-ios/releases/tag/5.0.0
- Android v8: https://github.com/RevenueCat/purchases-android/releases/tag/8.0.0

### DIFF
--- a/RevenueCat/Scripts/Purchases.cs
+++ b/RevenueCat/Scripts/Purchases.cs
@@ -131,6 +131,12 @@ public partial class Purchases : MonoBehaviour
                  || IsAndroidEmulator())
             apiKey = useAmazon ? revenueCatAPIKeyAmazon : revenueCatAPIKeyGoogle;
 
+        if (purchasesAreCompletedBy == PurchasesAreCompletedBy.MyApp && storeKitVersion == StoreKitVersion.Default)
+        {
+            Debug.Log("You must set a StoreKit version if you are setting PurchasesAreCompletedBy to MyApp. For Android, it doesn't matter which");
+            return;
+        }
+
         var dangerousSettings = new DangerousSettings(autoSyncPurchases);
         var builder = PurchasesConfiguration.Builder.Init(apiKey)
             .SetAppUserId(newUserId)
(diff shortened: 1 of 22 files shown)


tonidero:
Want to test this better but this should be ready for review

vegaro (approved):
Didn't see anything to call out

RevenueCat/Scripts/Purchases.cs
```diff
@@ -143,11 +134,11 @@ private void Configure(string newUserId)
         var dangerousSettings = new DangerousSettings(autoSyncPurchases);
         var builder = PurchasesConfiguration.Builder.Init(apiKey)
             .SetAppUserId(newUserId)
-            .SetObserverMode(observerMode)
+            .SetPurchasesAreCompletedBy(purchasesAreCompletedBy, storeKitVersion)
```
fire-at-will, line 137: 🤩🤩

fire-at-will (changes requested):
With the note that I'm _not_ at all familiar with Unity, all of the changes that you have here look good to me!

There is one function from iOS native 5.X that we should also bring in to be safe, [recordPurchase](https://github.com/RevenueCat/purchases-hybrid-common/blob/2a04578efaa6ff668c017b69f0224538c8118db4/ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift#L507). Let me know if you have any questions about the function and I'll be happy to help!

tonidero:
Ah yes thanks for the reminder! Will add that now

fire-at-will (approved):

fire-at-will:
The migration guide for Unity should be pretty similar to the react native one, just with Unity code samples instead of RN: https://github.com/RevenueCat/react-native-purchases/blob/main/migrations/v8-MIGRATION.md

RevenueCat/Scripts/Purchases.cs
```diff
@@ -140,14 +131,20 @@ private void Configure(string newUserId)
                  || IsAndroidEmulator())
             apiKey = useAmazon ? revenueCatAPIKeyAmazon : revenueCatAPIKeyGoogle;
 
+        if (purchasesAreCompletedBy == PurchasesAreCompletedBy.MyApp && storeKitVersion == StoreKitVersion.Default)
+        {
+            Debug.Log("You must set a StoreKit version if you are setting PurchasesAreCompletedBy to MyApp. For Android, it doesn't matter which");
```
tonidero, line 136: A couple things to note here, when configuring through the Editor:
- There is no breaking change even after removing the old observer mode. So if someone was setting observer mode in the editor before, there won't be any compilation issue when they migrate to the new version.
- There is also no compile-time check that makes it so you can't set "PurchasesAreCompletedBy.MyApp" without setting storeKitVersion as well.
For these 2 reasons, we're adding this runtime check which would make it so we don't configure the SDK in this situation and log this message.

PR #489, opened 2024-08-06
Update Deferred proration mode value to match BC7

tonidero:
With the BC7 update done in #485, we also had to update the Deferred mode underlying int value to match the new value expected by the billing client. This shouldn't have affected users since the previous change wasn't shipped yet


JayShortway (approved):
Confirmed that 6 is correct 😄

PR #500, opened 2024-09-05
Update VERSIONS.md to remove compatibility with Unity IAP for BC7

vegaro:
Unity IAP is not compatible with BC7


tonidero (approved):
Ahhh thanks for fixing that!

PR #539, opened 2024-12-19
Support web purchase Redemption Links

tonidero:
This adds support for Web Redemption Links in our Unity SDK.

PR #546, opened 2025-01-15
Fix Android SDK build issue

tonidero:
There was an issue building the SDK in android introduced in #539 and released in #545. 

I'm not sure why I didn't catch it when testing myself and it wasn't caught by the tests... I believe I might have messed up a rebase. In any case, I've tested the full redemption flow again with this change and it works as expected...

Thanks to @fire-at-will for noticing and suggesting the fix 🙏


JayShortway (approved):

vegaro (approved):

fire-at-will (commented):
👍

PR #549, opened 2025-01-21
Support Fetching and Redeeming Win-Back Offers on Custom Paywall

fire-at-will:
### Description 
This PR introduces four new functions to the Unity SDK that enable developers to fetch & redeem win-back offers on their custom paywalls. These functions are only supported on on iOS 18.0+ and require StoreKit 2 to be used.

#### Fetching eligible win-back offers
```c#
public delegate void GetEligibleWinBackOffersForProductFunc(WinBackOffer[] winBackOffers, Error error);
public void GetEligibleWinBackOffersForProduct(
   StoreProduct storeProduct, 
   GetEligibleWinBackOffersForProductFunc callback
)

public delegate void GetEligibleWinBackOffersForPackageFunc(WinBackOffer[] winBackOffers, Error error);
public void GetEligibleWinBackOffersForPackage(
   Package package, 
   GetEligibleWinBackOffersForPackageFunc callback
)
```

#### Redeeming win-back offers
```typescript
public void PurchaseProductWithWinBackOffer(
   StoreProduct storeProduct, 
   WinBackOffer winBackOffer, 
   MakePurchaseFunc callback
)

public void PurchasePackageWithWinBackOffer(
   Package package, 
   WinBackOffer winBackOffer, 
   MakePurchaseFunc callback
)
```

### Other Changes
The Subtester app was updated with a new buttons to support testing these flows

### Testing
All flows were tested manually through the new buttons in the Subtester app.


fire-at-will:
@RCGitBot please test

RevenueCat/Scripts/Purchases.cs
```diff
@@ -1271,6 +1271,72 @@ public void RedeemWebPurchase(WebPurchaseRedemption webPurchaseRedemption, Redee
...
+    private PurchaseProductWithWinBackOfferFunc PurchaseProductWithWinBackOfferCallback { get; set; }
+
+    /// <summary>
+    /// Purchase a product with a win-back offer. Only available on iOS 18.0+ with StoreKit 2.
+    /// Returns an error if the platform is not iOS 18.0+ or if StoreKit 2 is not used.
+    /// </summary>
+    /// <param name="storeProduct">The Purchases.StoreProduct to purchase</param>
+    /// <param name="winBackOffer">The Purchases.WinBackOffer to use</param>
+    /// <param name="callback">A callback receiving the product identifier, customer info, user cancellation, and an error if the purchase fails</param>
+    public void PurchaseProductWithWinBackOffer(StoreProduct storeProduct, WinBackOffer winBackOffer, PurchaseProductWithWinBackOfferFunc callback)
```
fire-at-will, line 1317: The other PurchaseProduct function accepts the `productIdentifier` as a string, but here I chose to accept the `StoreProduct` since that's more consistent with what we do in other places. 

Can y'all think of any reason that this might should accept the productIdentifier instead?

IntegrationTests/Assets/APITests/WinBackOfferAPITests.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        private void Start()
+        {
+            Purchases.WinBackOffer winBackOffer = new Purchases.WinBackOffer(null);
+            string identifier = winBackOffer.Identifier;
+            float price = winBackOffer.Price;
+            string priceString = winBackOffer.PriceString;
+            int cycles = winBackOffer.Cycles;
+            string period = winBackOffer.Period;
+            string periodUnit = winBackOffer.PeriodUnit;
```
tonidero, line 16: This is likely a bad positioning we had before, but I wonder if we should try to reuse the existing [Period](https://github.com/RevenueCat/purchases-unity/blob/5450caee72ce0532875367590d6ea1d85f70724c/RevenueCat/Scripts/SubscriptionOption.cs#L218) and [PeriodUnit](https://github.com/RevenueCat/purchases-unity/blob/5450caee72ce0532875367590d6ea1d85f70724c/RevenueCat/Scripts/SubscriptionOption.cs#L252) types

RevenueCat/Plugins/Android/PurchasesWrapper.java
```diff
@@ -596,6 +602,52 @@ public void onError(ErrorContainer errorContainer) {
...
+    public static void getEligibleWinBackOffersForProduct(String productIdentifier) {
+        // NOOP
+        PurchasesError error = new PurchasesError(
+                PurchasesErrorCode.UnsupportedError,
+                "Win-back offers are not supported on Android.");
+
+        ErrorContainer errorContainer = PurchasesErrorKt.map(
+                error,
+                new HashMap<>());
+        sendError(errorContainer, GET_ELIGIBLE_WIN_BACK_OFFERS_FOR_PRODUCT);
```
tonidero, line 614: I guess we could have moved some of this logic to PHC like we do for getPromotionalOffer:
```
ErrorContainer errorContainer = CommonKt.getPromotionalOffer();
        sendError(errorContainer, GET_PROMOTIONAL_OFFER);
```
But probably not worth changing it right now 👍

RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
```diff
@@ -442,6 +445,114 @@ - (void)showInAppMessages:(NSArray<NSNumber*>*)messageTypes {
...
+- (void)getEligibleWinBackOffersForProduct:(NSString *)productIdentifier {
+    // Checking for win-back offer APIs being available in the current OS is handled at the PHC level.
+    [RCCommonFunctionality eligibleWinBackOffersForProductIdentifier:productIdentifier
+                                                        completionBlock:^(NSArray<NSDictionary *> * _Nullable eligibleWinBackOffers, RCErrorContainer * _Nullable errorContainer) {
+
+        RCErrorContainer *finalErrorContainer = errorContainer;
+        if (eligibleWinBackOffers == nil && errorContainer == nil) {
+            NSError *nsError = [[NSError alloc] initWithDomain:RCPurchasesErrorCodeDomain
+                                                            code:RCUnknownError
+                                                        userInfo:@{NSLocalizedDescriptionKey: @"Both error and response are nil"}];
```
tonidero, line 457: I believe there are some indentation issues in this file

RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
```diff
@@ -442,6 +445,114 @@ - (void)showInAppMessages:(NSArray<NSNumber*>*)messageTypes {
...
+
+        // Send response dictionary if present
+        NSArray *offers = eligibleWinBackOffers ?: @[];
+        NSDictionary *response = @{@"eligibleWinBackOffers": offers};
+        [self sendJSONObject:response toMethod:GET_ELIGIBLE_WIN_BACK_OFFERS_FOR_PRODUCT];
+    }];
+}
+
+// This function accepts a product identifier since the PHC code only fetches eligible win-back offers for products
+- (void)getEligibleWinBackOffersForPackage:(NSString *)productIdentifier {
```
tonidero, line 476: Hmm I'm wondering if we need this method at all, and we should just have `getEligibleWinBackOffersForProduct`, and have the C# side only call that one internally... Wdyt?

RevenueCat/Scripts/Purchases.cs
```diff
@@ -1271,6 +1272,64 @@ public void RedeemWebPurchase(WebPurchaseRedemption webPurchaseRedemption, Redee
...
+    /// </summary>
+    /// <param name="storeProduct">The Purchases.StoreProduct to get win-back offers for</param>
+    /// <param name="callback">A callback receiving an array of Purchases.WinBackOffer objects or an error if unsuccessful</param>
+    public void GetEligibleWinBackOffersForProduct(StoreProduct storeProduct, GetEligibleWinBackOffersForProductFunc callback)
+    {
+        GetEligibleWinBackOffersForProductCallback = callback;
+        _wrapper.GetEligibleWinBackOffersForProduct(storeProduct);
+    }
+
+    public delegate void GetEligibleWinBackOffersForPackageFunc(WinBackOffer[] winBackOffers, Error error);
```
tonidero, line 1291: We could potentially have a single delegate for both methods in the API? Considering they return the same data.

Subtester/Assets/Scripts/PurchasesListener.cs
```diff
@@ -769,6 +773,199 @@ void ShowInAppMessages()
         Purchases.InAppMessageType.PriceIncreaseConsent, Purchases.InAppMessageType.Generic, Purchases.InAppMessageType.WinBackOffer });
     }
 
+    void PurchaseProductForWinBackTesting()
```
tonidero, line 776: Hmm what's the difference between this one and the existing purchase methods? Same for `PurchasePackageForWinBackTesting`

Subtester/SKConfig.storekit
```diff
@@ -0,0 +1,129 @@
+{
```
tonidero, line 1: Just to check but this file is included when exporting the project to iOS right?

tonidero (approved):
Some comments but nothing really blocking. The main question is whether we can simplify the native APIs by calling a single method with the appropriate parameters from the C# side.

Subtester/Assets/Scripts/PurchasesListener.cs
fire-at-will: The only difference is that it specifically purchases the product that has a win-back offer associated with it in the SKConfig file! If you purchase a different product, it won't have a win-back offer on it and you won't be able to test the functionality.

RevenueCat/Scripts/Purchases.cs
fire-at-will: Yeah, we probably could! I only kept them separate just in case they have to return different things in the future 🤷‍♂️

RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
fire-at-will: Good catch, fixed in d3ff240f8dd0b44f1260bc2a12b4bccd7f437083

IntegrationTests/Assets/APITests/WinBackOfferAPITests.cs
fire-at-will: Yeah, that's a great callout! The values are here like this to match what we're doing in purchase-unity's [Discount](https://github.com/RevenueCat/purchases-unity/blob/main/RevenueCat/Scripts/Discount.cs) class, and because it matches the interface used in the hybrids (which keep them separate because that's what the Discount class does in the other hybrids as well)

RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
fire-at-will: Yeah, we can probably get away with just `getEligibleWinBackOffersForProduct`. I'll work on making that change 💪

Subtester/SKConfig.storekit
fire-at-will: Right now it isn't, once Unity creates the xcworkspace, you need to manually add it. I added logs to the Subtester to remind people of this. 

Ideally we could add it to the xcworkspace that Unity generates, but I wasn't able to find a way to do it - do you know if that's possible?

Subtester/Assets/Scripts/PurchasesListener.cs
tonidero: Right... then I would say we should probably do a small refactor so only the product id becomes a parameter

RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
fire-at-will: Ah, I remember why I kept them separate now 😅

The native code here in `PurchasesUnityHelper.m` determines which function it sends the JSON response to:

```objc
[self sendJSONObject:response toMethod:GET_ELIGIBLE_WIN_BACK_OFFERS_FOR_PRODUCT];
```

I think the only way this would work is if `PurchasesUnityHelper.getEligibleWinBackOffersForProduct` took in the method name as an argument, which clouds up the API there instead. I felt like it was clearer to keep them as separate functions instead of adding that extra parameter there

tonidero: Ahh that makes sense yeah... I guess we could try to add another level of indirection in the C# side to avoid that... But that's going to complicate things and not follow the current architecture, so I think this is better. Thanks 🙏

Subtester/SKConfig.storekit
tonidero: I actually don't know either. I don't think this is a blocker, so I think we can go ahead with this for now 👍

tonidero (approved):

Subtester/Assets/Scripts/PurchasesListener.cs
fire-at-will: Did a small refactor so that these functions just call the existing purchase functions under the hood in  485633996fa3daf29ffd04344108c840ca8e0ed2 😄

RevenueCat/Scripts/Purchases.cs
```diff
@@ -1592,6 +1651,104 @@ private void _redeemWebPurchase(string redeemWebPurchaseJSON)
...
+
+        if (ResponseHasError(response))
+        {
+            MakePurchaseCallback(null, null, response["userCancelled"], new Error(response["error"]));
+        }
+        else
+        {
+            var info = new CustomerInfo(response["customerInfo"]);
+            var packageIdentifier = response["packageIdentifier"];
+            MakePurchaseCallback(packageIdentifier, info, false, null);
```
vegaro, line 1746: The `MakePurchaseCallback` first parameter is a `productIdentifier`, and you're passing the `packageIdentifier` here. I was looking at the implementation of PHC and it looks like in both `purchasePackage` and `purchaseProduct` we send the `productIdentifier` (you should double check but that's what I understand), so I think this should be sending the productIdentifier instead

vegaro (changes requested):

RevenueCat/Scripts/Purchases.cs
fire-at-will: Great catch!! 💜

fire-at-will: Fixed in e4a03d765b7ccbebb24253aa031ebbfaf53b21d0!

vegaro (approved):

PR #567, opened 2025-03-04
Add `subscriptionsByProductIdentifier` in CustomerInfo

tonidero:
This adds some data by subscription in the `CustomerInfo` that will allow to access some data previously unaccessible.

### DIFF
--- a/RevenueCat/Scripts/SubscriptionInfo.cs
+++ b/RevenueCat/Scripts/SubscriptionInfo.cs
@@ -1,0 +1,88 @@
+using UnityEngine;
+using System;
+using RevenueCat.SimpleJSON;
+using static RevenueCat.Utilities;
+
+
+public partial class Purchases
+{
+    
+    /// <summary>
+    /// The SubscriptionInfo object gives you access to all of the information about the status of a subscription.
+    /// </summary>
+    public class SubscriptionInfo
+    {
+        public readonly string ProductIdentifier;
+        public readonly DateTime PurchaseDate;
+        public readonly DateTime? OriginalPurchaseDate;
+        public readonly DateTime? ExpiresDate;
+        public readonly string Store;
+        public readonly DateTime? UnsubscribeDetectedAt;
+        public readonly bool IsSandbox;
+        public readonly DateTime? BillingIssuesDetectedAt;
+        public readonly DateTime? GracePeriodExpiresDate;
+        public readonly string OwnershipType;
+        public readonly string PeriodType;
+        public readonly DateTime? RefundedAt;
+        public readonly string? StoreTransactionId;
+        public readonly bool IsActive;
+        public readonly bool WillRenew;
+
+        public SubscriptionInfo(JSONNode response)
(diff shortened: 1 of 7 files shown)


JayShortway (approved):
It looks good to me! But it's perhaps wise to wait for approval of someone with more C# context. 🔈

RevenueCat/Scripts/SubscriptionInfo.cs
```diff
@@ -0,0 +1,88 @@
...
+            if (purchaseDateTime == null) {
+                Debug.LogError("Purchase date is null or has an invalid format. Defaulting to 1970-01-01.");
+            }
+            PurchaseDate = purchaseDateTime ?? new DateTime(1970, 1, 1);
+            OriginalPurchaseDate = FromResponseISO8601String(response, "originalPurchaseDate");
+            ExpiresDate = FromResponseISO8601String(response, "expiresDate");
+            Store = response["store"];
+            UnsubscribeDetectedAt = FromResponseISO8601String(response, "unsubscribeDetectedAt");
+            IsSandbox = response["isSandbox"].AsBool;
+            BillingIssuesDetectedAt = FromResponseISO8601String(response, "billingIssueDetectedAt");
```
vegaro, line 44: It looks like the  mapper maps it to `billingIssuesDetectedAt`, with an `s`

RevenueCat/Scripts/SubscriptionInfo.cs
```diff
@@ -0,0 +1,88 @@
...
+            GracePeriodExpiresDate = FromResponseISO8601String(response, "gracePeriodExpiresDate");
+            OwnershipType = response["ownershipType"];
+            PeriodType = response["periodType"];
```
vegaro, line 47: Should we have enums for `OwnershipType` and `PeriodType`?

vegaro (commented):
I think `billingIssuesDetectedAt` is broken, other than that it looks good

RevenueCat/Scripts/SubscriptionInfo.cs
tonidero: Good question... So `PeriodType` is used in `EntitlementInfo` and doesn't have it. That + us deciding not to use enums made me think to avoid it... Not an expert in C# to know what would be the best alternative though 😓. Wdyt?

RevenueCat/Scripts/SubscriptionInfo.cs
tonidero: Fixed in https://github.com/RevenueCat/purchases-unity/pull/567/commits/ce00afce2c0f6409c6f991360cf8cd6aceb9a7c9

RevenueCat/Scripts/SubscriptionInfo.cs
vegaro: if `EntitlementInfo` doesn't have it, then I think we are good!

vegaro (approved):

PR #586, opened 2025-04-17
Add `getStorefront` API

tonidero:
This adds a new API `Purchases.getStorefront` that returns an object of type `Storefront` which currently contains the country code to which the current Play store/App store/Amazon store account belongs to. In case there is an error fetching it, it will return null.

### DIFF
--- a/RevenueCat/Scripts/Purchases.cs
+++ b/RevenueCat/Scripts/Purchases.cs
@@ -221,6 +221,7 @@ public partial class Purchases : MonoBehaviour
 
     /// <summary>
     /// Fetches the <c>Storefront</c> for the customer's current store account.
+    /// If there is an error, the callback will be called with a null value.
     /// </summary>
     public void GetStorefront(GetStorefrontFunc callback)
     {
--- a/RevenueCat/Scripts/Purchases.cs
+++ b/RevenueCat/Scripts/Purchases.cs
@@ -212,6 +212,23 @@ public partial class Purchases : MonoBehaviour
     }
 
     /// <summary>
+    /// Callback type for <see cref="Purchases.GetStorefront"/>.
+    /// Includes the info of the current store account.
+    /// </summary>
+    public delegate void GetStorefrontFunc(Storefront? storefront);
+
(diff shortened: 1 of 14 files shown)


RevenueCat/Scripts/Storefront.cs
```diff
@@ -0,0 +1,25 @@
...
+        public Storefront(string countryCode)
+        {
+            CountryCode = countryCode;
+        }
+
+        public override string ToString()
+        {
+            return $"{nameof(CountryCode)}: {CountryCode}";
+        }
+    }
```
JayShortway, line 24: Just curious why this wrapper object exists. Do we expect to return more than just the country in the near future?

RevenueCat/Scripts/Purchases.cs
```diff
@@ -211,6 +211,24 @@ private bool IsAndroidEmulator()
...
+    /// Callback type for <see cref="Purchases.GetStorefront"/>.
+    /// Includes the info of the current store account.
+    /// </summary>
+    public delegate void GetStorefrontFunc(Storefront? storefront);
+
+    private GetStorefrontFunc StorefrontCallback { get; set; }
+
+    /// <summary>
+    /// Fetches the <c>Storefront</c> for the customer's current store account.
+    /// If there is an error, the callback will be called with a null value.
```
JayShortway, line 224: 🙌

JayShortway (approved):
Looks good! Just a question.

RevenueCat/Scripts/Storefront.cs
tonidero: Not in the near future... But I think it's conceivable we might want to add other data in the future. For example, country name (not code). Returning an object like this would make that much easier (as I've learned in the past 😅), and it's also the same API we have in iOS. The only different right now is Android, where I added a `getStorefrontCountryCode`, which could be deprecated in favor of `getStorefront` in the future if needed (I thought of going this route when trying to unify both the iOS and android APIs in PHC 😓  ).

JayShortway: Thanks for the context!

PR #605, opened 2025-06-19
Fix issue where chained calls to the same methods would cause the inner callback not to be called.

tonidero:
This is a possible approach to fix an issue where calling a method within the completion callback of that same method, would cause the listener reference to be cleared, and the inner callback not being called.

We might want to improve this to support concurrency and a safer system, but this fixes the issue for now.

### DIFF
--- a/RevenueCat/Scripts/Purchases.cs
+++ b/RevenueCat/Scripts/Purchases.cs
@@ -1353,10 +1353,12 @@ public partial class Purchases : MonoBehaviour
         Debug.Log("_receiveStorefront " + storefrontJson);
 
         if (StorefrontCallback == null) return;
+        var callback = StorefrontCallback;
+        StorefrontCallback = null;
 
         if (storefrontJson == null || storefrontJson == "{}")
         {
-            StorefrontCallback(null);
+            callback(null);
         }
         else
         {
@@ -1365,15 +1367,13 @@ public partial class Purchases : MonoBehaviour
             if (countryCode == null)
             {
                 Debug.LogError("StorefrontCallback received null countryCode");
-                StorefrontCallback(null);
+                callback(null);
             }
             else 
             {
-                StorefrontCallback(new Storefront(countryCode));
+                callback(new Storefront(countryCode));
             }
         }
-
-        StorefrontCallback = null;
     }
 
     // ReSharper disable once UnusedMember.Local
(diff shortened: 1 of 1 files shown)


RevenueCat/Scripts/Purchases.cs
```diff
@@ -1353,10 +1353,12 @@ private void _receiveStorefront(string storefrontJson)
         Debug.Log("_receiveStorefront " + storefrontJson);
 
         if (StorefrontCallback == null) return;
+        var callback = StorefrontCallback;
+        StorefrontCallback = null;
```
vegaro, line 1357: is this needed? the real fix is keeping that local `callback` reference in case other call overrides it right? I don't think it breaks anything but just curious

JayShortway (approved):
Nice pragmatic fix to support sequential calls! 💪

RevenueCat/Scripts/Purchases.cs
tonidero: Hmm we could have a memory leak if we don't clear the listener I guess? Also, note that the real fix is NOT clearing it after the callback is called. Because if during the callback, you call the same method, the listener gets set to the new listener, so we don't want to clear it after that.

RevenueCat/Scripts/Purchases.cs
```diff
@@ -1353,10 +1353,12 @@ private void _receiveStorefront(string storefrontJson)
         Debug.Log("_receiveStorefront " + storefrontJson);
 
         if (StorefrontCallback == null) return;
```
vegaro, line 1355: one thing I just thought of... we should probably move this check right after setting the callback, since it's possible it could have changed?

```
var callback = StorefrontCallback;
if (callback == null) return;
```

tonidero: Hmm if there is concurrency you're right... it won't work either way, but at least it would avoid some weird scenarios. Will change that

vegaro: since this is focusing on fixing sequential calls it can be done later as well, just something I thought of

RevenueCat/Scripts/Purchases.cs
vegaro: I see, makes sense!

vegaro: Another approach is to leave the clearing where it was but checking if it's the same as the saved callback:

```
if (ProductsCallback == callback)
{
    ProductsCallback = null;
}
```

I think that's also safer for concurrent calls, because this could potentially clear a brand new `StorefrontCallback` set concurrently right after setting the `var callback`

```
var callback = StorefrontCallback;
StorefrontCallback = null;
 ```
 
 But again, that's more related to concurrency, so feel free to ignore for this fix

tonidero: Right, even with that solution, we would still need to add some sort of locking mechanism to avoid it being changed between the equality check and setting it to null in any case. I think I will leave the current solution for now.

PR #625, opened 2025-07-18
Change result of purchase methods to PurchaseResult

tonidero:
This changes the result type of the purchase methods. Instead of returning a func with different parameters, we now return an object containing all the fields, which can make it more extensible.


ajpallares (approved):
Not my expertise, but it all made sense 👍

PR #634, opened 2025-07-28
[WIP] Virtual Currency Support

fire-at-will:
WIP

### DIFF
--- a/RevenueCat/Scripts/VirtualCurrencies.cs
+++ b/RevenueCat/Scripts/VirtualCurrencies.cs
@@ -1,5 +1,6 @@
 using System.Collections.Generic;
 using RevenueCat.SimpleJSON;
+using static RevenueCat.Utilities;
 
 public partial class Purchases
 {
--- a/RevenueCat/Scripts/VirtualCurrencies.cs
+++ b/RevenueCat/Scripts/VirtualCurrencies.cs
@@ -1,0 +1,30 @@
+using System.Collections.Generic;
+using RevenueCat.SimpleJSON;
+
+public partial class Purchases
+{
+    /// <summary>
+    /// The VirtualCurrencies object contains all the virtual currencies associated to the user.
+    /// </summary>
+    public class VirtualCurrencies
+    {
+        /// <summary>
+        /// Map of all VirtualCurrency objects keyed by virtual currency code.
+        /// </summary>
+        public readonly Dictionary<string, VirtualCurrency> All;
+
+        public VirtualCurrencies(JSONNode response)
+        {
+            All = new Dictionary<string, VirtualCurrency>();
+            foreach (var keyValuePair in response["all"])
+            {
+                All.Add(keyValuePair.Key, new VirtualCurrency(keyValuePair.Value));
+            }
+        }
+
+        public override string ToString()
+        {
+            return $"{nameof(All)}:\n{DictToString(All)}";
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


RevenueCat/Scripts/Purchases.cs
```diff
@@ -1286,6 +1286,34 @@ public void RedeemWebPurchase(WebPurchaseRedemption webPurchaseRedemption, Redee
         _wrapper.RedeemWebPurchase(webPurchaseRedemption);
     }
 
+    public delegate void GetVirtualCurrenciesFunc(VirtualCurrencies? virtualCurrencies, Error error);
```
fire-at-will: Here, I've marked `virtualCurrencies` as optional (with `VirtualCurrencies?`) to reflect that it can be null if an error occurs. This doesn't seem to be the convention in this file though, other functions seem to omit the `?` in both the return object _and_ the optional error.

Do y'all think it's better to include the `?` here, or stick to the conventions in the API and make this just be a `VirtualCurrencies` type?

Same goes for the `public VirtualCurrencies? GetCachedVirtualCurrencies()` function below, except there, the cached virtual currencies can be null, even if no error occurs.

RevenueCat/Scripts/VirtualCurrencies.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    /// <summary>
+    /// The VirtualCurrencies object contains all the virtual currencies associated to the user.
+    /// </summary>
+    public class VirtualCurrencies
+    {
+        /// <summary>
+        /// Map of all VirtualCurrency objects keyed by virtual currency code.
+        /// </summary>
+        public readonly Dictionary<string, VirtualCurrency> All;
```
tonidero, line 15: Should we use a `ReadOnlyDictionary` as the public API? If I understand correctly, this would allow developers to modify the dictionary after they obtain it and we might want to avoid that.

I think we haven't used that before... But feels like we should 😅

RevenueCat/Scripts/Purchases.cs
tonidero: That's a great point! So from what I read, nullable reference types have been enabled by default in C# for a while but seems that Unity doesn't have them enabled... So basically adding the `?` would just be a bit of an indicator, but is not enforced even if not added...

I do think it's probably worth to start making the types nullable since it seems more correct. Great catch!!

In this case, we should also make `Error` nullable right?

tonidero (approved):
Just a couple comments, but looking great!! Amazing job finding that issue with nullables!

RevenueCat/Scripts/VirtualCurrencies.cs
fire-at-will: Do we have a minimum supported Unity version that we support, or are all Unity versions supported by the SDK? 

[ReadOnlyDictionary's docs](https://docs.unity3d.com/Packages/com.unity.reflect@1.1/api/Unity.Labs.Utils.ReadOnlyDictionary-2.html) seem to suggest that ReadOnlyDictionary is available in .NET 4.5+, and is polyfilled "under .Net 3.5". It's not super clear to me whether this means that the polyfill is available in .NET versions <3.5, or only in .NET versions >=3.5

tonidero: To be honest, I'm not sure, but that's a great point! I hadn't considered that... So to avoid any issues and considering we're already using Dictionary in other places, let's keep it as is for now. Thanks!

MarkVillacampa (approved):

PR #639, opened 2025-08-07
Add pricePerWeek/Month/Year convenience properties to StoreProduct

tonidero:
Adds new `pricePerWeek`, `pricePerMonth`, `pricePerYear`, `pricePerWeekString`, `pricePerMonthString`, `pricePerYearString` properties to `StoreProduct

### DIFF
--- a/RevenueCat/Scripts/StoreProduct.cs
+++ b/RevenueCat/Scripts/StoreProduct.cs
@@ -43,40 +43,58 @@ public partial class Purchases
         public readonly string PriceString;
 
         /// <summary>
-        /// Price of the product per week. Null for non-subscription products.
+        /// Null for non-subscription products. The price of the StoreProduct in a weekly recurrence.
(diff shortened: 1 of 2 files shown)


RevenueCat/Scripts/StoreProduct.cs
```diff
@@ -42,6 +42,45 @@ public class StoreProduct
         /// <returns></returns>
         public readonly string PriceString;
 
+        /// <summary>
+        /// Price of the product per week. Null for non-subscription products.
```
JayShortway: Should we clarify that these are approximations? I believe we do that in some hybrids already.

tonidero: Good call! I just copied the docs from `purchases-hybrid-common` typescript interfaces which seemed more complete.

ajpallares (approved):

PR #649, opened 2025-08-28
Paywalls

vegaro:
We created a `PaywallProxyActivity`, instead of using the regular fragment we normally use for presenting paywalls in the hybrids. The reason is that we can't use Unity apps use UnityPlayerActivity as the main activity, which is not FragmentActivity. In Flutter we ask devs to implement FragmentActivity in their Activity, but it's not easy to do that in Unity.

For callbacks we use UnitySendMessage, since that's the simplest way without having to rewrite the way we present paywalls to be able to communicate paywall results. It Unity's built-in GameObject messaging system, which I tested and acts as a queue, which is nice since it means the message will come back as soon as `PaywallProxyActivity` is closed, behaving very similar to `onActivityResult`. This avoids rewriting existing paywall presentation architecture, which was in progress and we can still do if we want to support real-time paywall notifications.

To hold the Java code, I created a `androidlib` folder, following the docs https://docs.unity3d.com/Manual/android-library-plugin-create.html. The reason is that it's the only way to include an AndroidManifest.xml, the alternative is building an aar, which wuld require external build process and complicate development workflow.

**EDIT: The circular dependency issue is not an issue anymore and has been solved with https://github.com/RevenueCat/purchases-unity/pull/649#issuecomment-3337204781, but keeping it here for reference**

## Circular dependency issue
This is how the Gradle dependencies look like:

```
unityLibrary (main) → depends on → RevenueCatUI.androidlib (our plugin)
RevenueCatUI.androidlib → CANNOT depend on → unityLibrary (circular!)
```

Screenshot 2025-09-22 at 13 30 34

We need to call Unity functions (see `UnityBridge`) using reflection. The androidlib is converted into a module, and the `unityLibrary` module depends on it (`implementation project(':RevenueCatUI.androidlib'`). That means adding a dependency on `unityLibrary` in `RevenueCatUI.androidlib` would create a circular dependency. That's why I build the `UnityBridge`.

### Usage

```
using RevenueCat.UI;

// Basic paywall presentation
var ui = GetComponent<RevenueCatUI>();
var result = await ui.PresentPaywall();

// Paywall with options
var options = new PaywallOptions {
    OfferingIdentifier = "premium_monthly",
    DisplayCloseButton = true
};
var result = await ui.PresentPaywall(options);

// Conditional paywall
var result = await ui.PresentPaywallIfNeeded("premium_features", options);
```

### Followups
- Font support
- PresentedOfferingContext support


RevenueCat/Plugins/Android/PurchasesWrapper.java
```diff
@@ -101,6 +105,21 @@ public static void setup(String apiKey,
...
     }
 
+    public static void presentPaywallForResult(@Nullable String offeringIdentifier) {
+        try {
+            android.app.Activity activity = UnityPlayer.currentActivity;
+            Intent intent = new Intent(activity, com.revenuecat.purchasesunity.PaywallProxyActivity.class);
+            intent.putExtra(com.revenuecat.purchasesunity.PaywallProxyActivity.EXTRA_GAME_OBJECT, gameObject);
+            intent.putExtra(com.revenuecat.purchasesunity.PaywallProxyActivity.EXTRA_METHOD, PAYWALL_RESULT);
+            if (offeringIdentifier != null) {
+                intent.putExtra(com.revenuecat.purchasesunity.PaywallProxyActivity.EXTRA_OFFERING_ID, offeringIdentifier);
```
vegaro: this is really being ignored atm

RevenueCat/Plugins/Android/AndroidManifest.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
```
tonidero, line 2: Is this file merged automatically on the Unity resulting AndroidManifest? Or do developers need to configure something to get it merged? 

Just asking in case it's worth moving this to PHC android, even if it's unused in the other hybrids...

RevenueCat/Plugins/Android/PaywallProxyActivity.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Override
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        
+        Intent source = getIntent();
+        gameObject = source.getStringExtra(EXTRA_GAME_OBJECT);
+        method = source.getStringExtra(EXTRA_METHOD);
+
+        PaywallActivityLauncher launcher = new PaywallActivityLauncher(this, result -> {
```
tonidero: Hmm would it be worth to keep a reference to the launcher at the activity level? I guess if we've tested and it works that it's not being released... but feels the proper pattern would be to hold a reference to the launcher, like in: https://github.com/RevenueCat/purchases-android/blob/7c41870f97c478b1ea20553217b11470dbda0ecd/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/MainActivity.kt#L25

RevenueCat/Plugins/Android/PaywallProxyActivity.java
```diff
@@ -0,0 +1,66 @@
...
+        Log.d("PurchasesUnity", "Launching paywall with PaywallActivityLauncher");
+        launcher.launch(null);
+    }
+
+    private void sendPaywallResult(PaywallResult result) {
+        if (gameObject == null || method == null) return;
+        
+        String resultName;
+        if (result instanceof PaywallResult.Purchased) {
+            resultName = "purchased";
```
tonidero: We should make sure this matches what we expect from iOS.

RevenueCat/Plugins/Android/PurchasesWrapper.java
```diff
@@ -101,6 +105,21 @@ public static void setup(String apiKey,
         Purchases.getSharedInstance().setUpdatedCustomerInfoListener(listener);
     }
 
+    public static void presentPaywallForResult(@Nullable String offeringIdentifier) {
```
tonidero: This is something we're currently working on, but we should also pass the `PresentedOfferingContext` to this method, so we can keep track of this information. We still need to make this work in PHC (natives done, iOS PHC part is done [here](https://github.com/RevenueCat/purchases-hybrid-common/pull/1246), still missing android PHC part).

RevenueCat/Plugins/Android/PurchasesWrapper.java
```diff
@@ -101,6 +105,21 @@ public static void setup(String apiKey,
...
+        try {
+            android.app.Activity activity = UnityPlayer.currentActivity;
```
tonidero: We should import instead of using the fully qualified name?  Have a bunch like this in this file.

RevenueCat/Plugins/Android/PurchasesWrapper.java
```diff
@@ -101,6 +105,21 @@ public static void setup(String apiKey,
...
+            Intent intent = new Intent(activity, com.revenuecat.purchasesunity.PaywallProxyActivity.class);
+            intent.putExtra(com.revenuecat.purchasesunity.PaywallProxyActivity.EXTRA_GAME_OBJECT, gameObject);
+            intent.putExtra(com.revenuecat.purchasesunity.PaywallProxyActivity.EXTRA_METHOD, PAYWALL_RESULT);
+            if (offeringIdentifier != null) {
+                intent.putExtra(com.revenuecat.purchasesunity.PaywallProxyActivity.EXTRA_OFFERING_ID, offeringIdentifier);
+            }
+            activity.startActivity(intent);
+        } catch (Throwable t) {
+            Log.e("Purchases", "Error launching PaywallProxyActivity", t);
```
tonidero: We wouldn't be returning any result to the Unity side callback in this case right? Not sure in what cases this could error though... I think maybe if the activity is finished, not sure if something else.

RevenueCat/Scripts/PaywallResult.cs
```diff
@@ -0,0 +1,64 @@
+public partial class Purchases
+{
+    /// <summary>
+    /// Possible values for the result of a paywall presentation.
+    /// Matches Flutter's PaywallResult enum for consistency across platforms.
+    /// </summary>
+    public enum PaywallResult
+    {
+        /// <summary>
+        /// Only returned when using presentPaywallIfNeeded. Returned if the paywall was not presented.
```
tonidero: We still haven't implemented `presentPaywallIfNeeded` in this PR right? I think it shouldn't be too difficult... though not sure if launching the proxy activity and closing it immediately could cause some weird UI glitch?

RevenueCat/Plugins/Android/AndroidManifest.xml
```diff
@@ -0,0 +1,17 @@
...
+          xmlns:tools="http://schemas.android.com/tools">
+
+    <application>
+        <activity
+            android:name="com.revenuecat.purchasesunity.PaywallProxyActivity"
+            android:exported="false"
+            android:theme="@style/Theme.AppCompat.NoActionBar"
```
tonidero: This is related to the `presentPaywallIfNeeded`, but I wonder if we need to make this activity a "transparent" activity to better support `presentPaywallIfNeeded`...

RevenueCat/Plugins/Android/PaywallProxyActivity.java
```diff
@@ -0,0 +1,66 @@
...
+
+import androidx.annotation.Nullable;
+import androidx.appcompat.app.AppCompatActivity;
+
+import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallActivityLauncher;
+import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallResult;
+import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallResultHandler;
+import com.unity3d.player.UnityPlayer;
+
+public class PaywallProxyActivity extends AppCompatActivity {
```
tonidero: I wonder if we should use `ComponentActivity` instead... I think that would have what we need, without needing to import the whole appcompat library? I think we might still need the [androidx.activity](https://developer.android.com/jetpack/androidx/releases/activity) dependency

RevenueCat/Plugins/Android/AndroidManifest.xml
vegaro: it gets merged automatically in the developers apps. Since it's placed the `Plugins` folder. I think it's easier to keep it here until we need it in another hybrid, it makes local developing way easier. Also the proxy activity is sending Unity messages directly atm

RevenueCat/Plugins/Android/PurchasesWrapper.java
vegaro: oh yes oopsies

RevenueCat/Scripts/PaywallResult.cs
vegaro: no, I didn't add any parameter yet, just wanted to make sure the approach made sense. But your glitch point makes sense. Let me test it out.

RevenueCat/Plugins/Android/PaywallProxyActivity.java
vegaro: Good idea!

RevenueCat/Plugins/Android/PurchasesWrapper.java
vegaro: Added as a followup. This and fonts support

RevenueCat/Plugins/Android/AndroidManifest.xml
vegaro: I checked and didn't see any glitch or anything

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/UnityBridge.java
```diff
@@ -0,0 +1,28 @@
+package com.revenuecat.purchasesunity.ui;
+
+import android.app.Activity;
+import android.util.Log;
+
+final class UnityBridge {
```
facumenzella: This deserves a comment explaining what's in the PR description

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/RevenueCatUI.java
```diff
@@ -0,0 +1,98 @@
+package com.revenuecat.purchasesunity.ui;
+
+import static com.revenuecat.purchasesunity.ui.PaywallProxyActivity.EXTRA_GAME_OBJECT;
+import static com.revenuecat.purchasesunity.ui.PaywallProxyActivity.EXTRA_METHOD;
+import static com.revenuecat.purchasesunity.ui.PaywallProxyActivity.EXTRA_OFFERING_ID;
+import static com.revenuecat.purchasesunity.ui.PaywallProxyActivity.EXTRA_SHOULD_DISPLAY_DISMISS_BUTTON;
```
facumenzella: I see that in some places we call this `DISMISS_BUTTON`, but in other places we have a `CloseButton`. Let's use the same naming, either dismiss or close

RevenueCatUI/Scripts/RevenueCatUI.cs
```diff
@@ -7,24 +7,24 @@ namespace RevenueCat.UI
     /// <summary>
     /// Main interface for RevenueCat UI components.
     /// Provides methods to present paywalls.
+    /// Add this component to a GameObject to enable paywall functionality.
     /// </summary>
-    public static class RevenueCatUI
+    public class RevenueCatUI : MonoBehaviour
```
facumenzella:
```suggestion
    public sealed class RevenueCatUI : MonoBehaviour
```

I am not 100% sure about it, but feels like this class shouldn't be subclassed

Subtester/Packages/manifest.json
```diff
@@ -1,6 +1,7 @@
 {
   "dependencies": {
     "com.revenuecat.purchases-unity": "file:../../RevenueCat",
+    "com.revenuecat.purchases-ui-unity": "file:../../RevenueCatUI",
```
facumenzella, line 4: is this safe? 👀

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/AndroidManifest.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+          xmlns:tools="http://schemas.android.com/tools">
+
+    <application>
+        <activity
+            android:name=".PaywallProxyActivity"
+            android:exported="false"
+            android:theme="@android:style/Theme.NoTitleBar"
+            tools:node="merge" />
```
JayShortway, line 10: These kinds of activities are often called "trampoline" activities. Maybe we can use that name for ours as well, `PaywallTrampolineActivity`?

[This blog post](https://marcellogalhardo.dev/posts/2023/trampoline-activities/) by Marcello Galhardo has some details on the pattern, one of which is the tip to use the `@android:style/Theme.NoDisplay` theme. We discussed that we should ensure no UI glitches and transition animations happen. Maybe that theme is all we need?

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/build.gradle
```diff
@@ -0,0 +1,49 @@
+apply plugin: 'com.android.library'
+
+dependencies {
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
+    implementation 'com.revenuecat.purchases:purchases-hybrid-common:17.7.0'
+    implementation 'com.revenuecat.purchases:purchases-hybrid-common-ui:17.7.0'
```
JayShortway: We should not forget to ensure these get bumped automatically.

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallTrampolineActivity.java
```diff
@@ -0,0 +1,134 @@
...
+                        sendPaywallResult(result);
+                    }
+                } finally {
+                    finish();
+                }
+            }
+        );
+
+        if (requiredEntitlementIdentifier != null) {
+            Log.d(TAG, "Using launchIfNeeded for entitlement '" + requiredEntitlementIdentifier + "'");
```
JayShortway: Just checking, but do we intend to keep all these log messages? At the moment they're not controlled by our `logLevel` configuration option.

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/RevenueCatUI.java
```diff
@@ -0,0 +1,98 @@
...
+            
+            if (offeringIdentifier != null && !offeringIdentifier.isEmpty()) {
+                intent.putExtra(EXTRA_OFFERING_ID, offeringIdentifier);
+            }
+            
+            intent.putExtra(EXTRA_SHOULD_DISPLAY_DISMISS_BUTTON, displayCloseButton);
+
+            Log.d(TAG, "Launching PaywallProxyActivity for gameObject=" + gameObject +
+                    ", offering=" + offeringIdentifier + ", displayCloseButton=" + displayCloseButton);
+            activity.startActivity(intent);
```
JayShortway: Could we give `PaywallProxyActivity` some static functions to do this in a type safe way, so callers don't have to worry about using the right keys for the right arguments? I'm thinking:
- `PaywallProxyActivity.presentPaywall()` 
- `PaywallProxyActivity.presentPaywallIfNeeded()` 

They both take the relevant arguments, build the intent, and run `startActivity()`.

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/RevenueCatUI.java
```diff
@@ -0,0 +1,98 @@
...
+        } catch (Throwable t) {
+            Log.e(TAG, "Error launching PaywallProxyActivity for presentPaywallIfNeeded", t);
+            UnityBridge.sendMessage(gameObject, "OnPaywallResultFromActivity", "ERROR|" + t.getClass().getSimpleName());
+        }
+    }
+
+    public static boolean isSupported() { return true; }
+
+    // Keeps your callback path intact if you use it internally (not used by SendMessage flow)
```
JayShortway: This comment is not needed I think 😄

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/UnityBridge.java
```diff
@@ -0,0 +1,28 @@
...
+    private static final String TAG = "UnityBridge";
+    private UnityBridge() {}
+
+    static Activity currentActivityOrNull() {
+        try {
+            Class<?> up = Class.forName("com.unity3d.player.UnityPlayer");
```
JayShortway: Just checking, but is it not possible to depend _only_ on `com.unity3d.player`, without depending on the full `unityLibrary` thing?

RevenueCatUI/Plugins/Editor/RevenueCatUIDependencies.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<dependencies>
+  <androidPackages>
+    <androidPackage spec="com.revenuecat.purchases:purchases-hybrid-common-ui:[17.7.0]" />
+    <androidPackage spec="androidx.activity:activity:1.8.2" />
+  </androidPackages>
+  
+  <iosPods>
+    <iosPod name="PurchasesHybridCommonUI" version="~> 17.7.0" />
```
JayShortway: Just noting that these should also be kept up to date automatically.

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallTrampolineActivity.java
JayShortway, line 1: Come to think of it, why are we not moving this to PHC? At that point we could even remove the intermediate `PaywallFragment`, right? In any case, using this trampoline pattern in PHC would also avoid the need for `FragmentActivity` in purchases-flutter.

vegaro: I thought about it but I don't know how we would do the `UnityBridge.sendMessage(gameObject, method, resultName));`, since we can't do `onActivityResult` directly from `RevenueCatUI.java`.

I think the only way would be to have a singleton in phc that we use as a message passing mechanism. 

Or am I missing anything?

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/AndroidManifest.xml
vegaro: Nice blogpost! It looks like "trampoline" activities are finished as soon as they are created:

```
class MyTrampolineActivity : ComponentActivity() {
  override fun onCreate(savedInstanceState: Bundle?) {
    super.onCreate(savedInstanceState)

    // Start your next activity.
    finish()
  }
}
```

But in our case, we want the Activity to send the message through the Unity bridge. If we finish the Activity we won't be able to send that. We need the activity to stay alive to receive PaywallResult from the launcher basically.

Also from [the docs](https://developer.android.com/reference/android/R.style.html#Theme_NoDisplay) of `Theme.NoDisplay`

> Default theme for activities that don't actually display a UI; that is, they finish themselves before being resumed.

I could create a static callback manager that implements `PaywallResultHandler` for the results of the PaywallActivity, but I think there's risk of leaking the trampoline Activity? What do you think? @JayShortway

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallTrampolineActivity.java
vegaro: No this is just for now, for debugging since it's a bit cumbersome to set breakpoints and such. I will add `TODO` around them to make sure we remove them.

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/UnityBridge.java
vegaro: Screenshot 2025-09-26 at 08 46 50

`com.unity3d.player.UnityPlayer` and other `com.unity3d.player` are packed in a `unity-classes.jar` inside the `unityLibrary` module.

One thing that's possible to do is making the **dependent** module the `launcher` module.

So right now this is what's set (the default) on the `.androidlib`:

Screenshot 2025-09-26 at 08 48 47

But `launcher` depends on `unityLibrary` already:

```
implementation project(':unityLibrary')
```

_I think_ we could do:

```
Launcher -> UnityLibrary
         \-> RevenueCatUI.androidlib -> UnityLibrary
```

Or even what's mentioned here which is not having any dependent module:

https://docs.unity3d.com/6000.3/Documentation/Manual/android-library-project-import.html

```
Launcher -> UnityLibrary
RevenueCatUI.androidlib -> UnityLibrary
```

I will give the latter a try because it feels like we should probably be doing that

vegaro: I tried setting `None` as the dependent module, and it's having issues finding the Java classes from C# in:

```
_plugin = new AndroidJavaClass("com.revenuecat.purchasesunity.ui.RevenueCatUI");
```

vegaro:
ok... I have good news. I have been experimenting with how we could remove the dependency on `unityLibrary` and I think I figured it out

First of all, I am very surprised about this, but it's possible to pass an activity over the bridge:

```c#
using (var unityPlayer = new AndroidJavaClass("com.unity3d.player.UnityPlayer"))
{
    var currentActivity = unityPlayer.GetStatic<AndroidJavaObject>("currentActivity");
    _plugin.CallStatic("presentPaywall", new object[] { currentActivity, offering, displayCloseButton });
}
```

The Java code gets it fine:

```java
public static void presentPaywall(Activity activity, String offeringIdentifier, boolean displayCloseButton). {
```

Then for the `UnitySendMessage` stuff. I actually had some leftover code already for this. We can have an static function in `RevenueCatUI.java`

```java
public interface PaywallCallbacks { void onPaywallResult(String result); }

private static volatile PaywallCallbacks paywallCallbacks;

public static void registerPaywallCallbacks(PaywallCallbacks cb) { paywallCallbacks = cb; }
public static void unregisterPaywallCallbacks() { paywallCallbacks = null; }

public static void sendPaywallResult(String result) {
    PaywallCallbacks cb = paywallCallbacks;
    cb.onPaywallResult(result);
}
```

Callbacks are registered from C#:

```c#
private class CallbacksProxy : AndroidJavaProxy
{
    private readonly AndroidPaywallPresenter _owner;
    public CallbacksProxy(AndroidPaywallPresenter owner) : base("com.revenuecat.purchasesunity.ui.RevenueCatUI$PaywallCallbacks")
    {
        _owner = owner;
    }

    // Signature matches Java interface
    public void onPaywallResult(string result)
    {
        _owner.OnPaywallResult(result);
    }
}

private readonly CallbacksProxy _callbacks;

public AndroidPaywallPresenter()
{
    _plugin = new AndroidJavaClass("com.revenuecat.purchasesunity.ui.RevenueCatUI");
    _callbacks = new CallbacksProxy(this);
    _plugin.CallStatic("registerPaywallCallbacks", _callbacks);
}

~AndroidPaywallPresenter()
{
    _plugin.CallStatic("unregisterPaywallCallbacks");
}
```

And then the owner of the PaywallCallbacks (`AndroidPaywallPresenter`) implements `OnPaywallResult`:

```c#
public void OnPaywallResult(string resultData)
{
    if (_current == null) return;
    try
    {
        var token = resultData?.Split('|')[0] ?? "ERROR";
        var type = PaywallResultTypeExtensions.FromNativeString(token);
        _current.TrySetResult(new PaywallResult(type));
    }
    catch (Exception e)
    {
        Debug.LogError($"[RevenueCatUI][Android] Failed to handle paywall result '{resultData}': {e.Message}. Setting Error.");
        _current.TrySetResult(PaywallResult.Error);
    }
    finally
    {
        _current = null;
    }
}
```

And then in `PaywallTrampolineActivity` we call the static method in `RevenueCatUI`:

```java
private void sendPaywallResult(PaywallResult result) {
    ...
    RevenueCatUI.sendPaywallResult(resultName);
}
```

I really like this since it doesn't rely on `UnitySendMessage`. I still don't know how we could move `PaywallTrampolineActivity` to phc since we would need a way to modify the "static" method the Activity should call to send the paywall results. We could use reflection and look for an interface but it's sketchy.

vegaro:
ok... I have good news. I have been experimenting with how we could remove the dependency on `unityLibrary` and I think I figured it out

First of all, I am very surprised about this, but it's possible to pass an activity over the bridge:

```c#
using (var unityPlayer = new AndroidJavaClass("com.unity3d.player.UnityPlayer"))
{
    var currentActivity = unityPlayer.GetStatic<AndroidJavaObject>("currentActivity");
    _plugin.CallStatic("presentPaywall", new object[] { currentActivity, offering, displayCloseButton });
}
```

The Java code gets it fine:

```java
public static void presentPaywall(Activity activity, String offeringIdentifier, boolean displayCloseButton). {
```

Then for the `UnitySendMessage` stuff. I actually had some leftover code already for this. We can have an static function in `RevenueCatUI.java`

```java
public interface PaywallCallbacks { void onPaywallResult(String result); }

private static volatile PaywallCallbacks paywallCallbacks;

public static void registerPaywallCallbacks(PaywallCallbacks cb) { paywallCallbacks = cb; }
public static void unregisterPaywallCallbacks() { paywallCallbacks = null; }

public static void sendPaywallResult(String result) {
    PaywallCallbacks cb = paywallCallbacks;
    cb.onPaywallResult(result);
}
```

Callbacks are registered from C#:

```c#
private class CallbacksProxy : AndroidJavaProxy
{
    private readonly AndroidPaywallPresenter _owner;
    public CallbacksProxy(AndroidPaywallPresenter owner) : base("com.revenuecat.purchasesunity.ui.RevenueCatUI$PaywallCallbacks")
    {
        _owner = owner;
    }

    // Signature matches Java interface
    public void onPaywallResult(string result)
    {
        _owner.OnPaywallResult(result);
    }
}

private readonly CallbacksProxy _callbacks;

public AndroidPaywallPresenter()
{
    _plugin = new AndroidJavaClass("com.revenuecat.purchasesunity.ui.RevenueCatUI");
    _callbacks = new CallbacksProxy(this);
    _plugin.CallStatic("registerPaywallCallbacks", _callbacks);
}

~AndroidPaywallPresenter()
{
    _plugin.CallStatic("unregisterPaywallCallbacks");
}
```

And then the owner of the PaywallCallbacks (`AndroidPaywallPresenter`) implements `OnPaywallResult`:

```c#
public void OnPaywallResult(string resultData)
{
    if (_current == null) return;
    try
    {
        var token = resultData?.Split('|')[0] ?? "ERROR";
        var type = PaywallResultTypeExtensions.FromNativeString(token);
        _current.TrySetResult(new PaywallResult(type));
    }
    catch (Exception e)
    {
        Debug.LogError($"[RevenueCatUI][Android] Failed to handle paywall result '{resultData}': {e.Message}. Setting Error.");
        _current.TrySetResult(PaywallResult.Error);
    }
    finally
    {
        _current = null;
    }
}
```

And then in `PaywallTrampolineActivity` we call the static method in `RevenueCatUI`:

```java
private void sendPaywallResult(PaywallResult result) {
    ...
    RevenueCatUI.sendPaywallResult(resultName);
}
```

I really like this since it doesn't rely on `UnitySendMessage`. I still don't know how we could move `PaywallTrampolineActivity` to phc since we would need a way to modify the "static" method the Activity should call to send the paywall results. We could use reflection and look for an interface but it's sketchy.

@facumenzella @JayShortway

JayShortway:
@vegaro This is great! Just thinking about how we can move `PaywallTrampolineActivity` to PHC (absolutely not a requirement), we might be able to keep the listener in a singleton (`PaywallResultListenerRegistry`)?

In PHC:
```kotlin
object PaywallResultListenerRegistry {
  var listener: PaywallResultListener? = null
}
```
```java
class PaywallTrampolineActivity {

  static void start(listener: PaywallResultListener) {
    PaywallResultListenerRegistry.listener = listener
  }

  private void sendPaywallResult(PaywallResult result) {
    // ...
    PaywallResultListener listener = PaywallResultListenerRegistry.listener
    if (listener != null) listener.onPaywallResult(result)
  }

}

```

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/AndroidManifest.xml
JayShortway: >We need the activity to stay alive

Hmm good point, but I don't think we can assume that actually. Depending on the device and memory constraints, activities on the backstack might be killed. We can test this by enabling [Don't keep activities](https://stackoverflow.com/a/22402360) in the developer options.

Maybe we need a callback manager after all? Why do you think there's a risk of leaking the trampoline Activity (just curious)?

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallTrampolineActivity.java
JayShortway: Hah, I wrote [this](https://github.com/RevenueCat/purchases-unity/pull/649#issuecomment-3337248489) before I saw this comment. Great minds and all that 😄

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/AndroidManifest.xml
vegaro: Hmm now I am actually not sure. My thinking was:

1. `launcher` is created in `PaywallTrampolineActivity`. The `PaywallResultHandler` is a static callback manager
1. `PaywallTrampolineActivity` is finished before `onResume`.
1. Does the `launcher` prevent `PaywallTrampolineActivity` from being garbage collected because it holds a reference to the `PaywallResultHandler` which is a static class that's not destroyed?

Like can the `PaywallResultHandler` live longer than the launcher?

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallTrampolineActivity.java
tonidero: Hmm so I'm not in love with holding references to objects like the listener statically... I think for most cases it would be ok, but could potentially lead to some memory leaks, and we should make sure the cleanup process works correctly... So in general I would try to avoid it, but considering that in Unity we kinda have to, I could see that we might want to move the logic to PHC in case it can be useful for other hybrids... 

IMO, I do prefer the current production pattern we use in the other hybrids, where we don't hold anything statically as long as it works. I would only use this static holder approach when there is no alternative. Do we have other use cases for other hybrids where we need this? if so, I would be ok moving it to PHC. If not, I think better to keep it to Unity, and leave the other hybrids use the existing approach.

vegaro: I think the main advantage would be to move Flutter to this system, so developers don't need to implement `FlutterFragmentActivity`

vegaro: But I kinda agree that I would prefer the `FlutterFragmentActivity` approach over the static `PaywallResultListenerRegistry` as well.

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallTrampolineActivity.java
```diff
@@ -0,0 +1,128 @@
...
+    private PaywallActivityLauncher launcher;
+
+    @Override
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        final Intent source = getIntent();
+        String offeringId = source.getStringExtra(EXTRA_OFFERING_ID);
+        boolean shouldDisplayDismissButton = source.getBooleanExtra(EXTRA_SHOULD_DISPLAY_DISMISS_BUTTON, false);
+        String requiredEntitlementIdentifier = source.getStringExtra(EXTRA_REQUIRED_ENTITLEMENT_IDENTIFIER);
```
tonidero: I think it could be nice to create a new data class type with all these parameters, so we can pass a single parcelable argument with all these parameters, and then we get that object, so it's easier to add/modify parameters that way, rather than having to add new individual parameters like this. Wdyt?

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallTrampolineActivity.java
```diff
@@ -0,0 +1,128 @@
...
+        Log.d(TAG, "Launching paywall if needed with PaywallActivityLauncher");
+        Log.d(TAG, "Options - entitlement: " + requiredEntitlementIdentifier + ", offering: " + offeringId + ", dismissButton: " + shouldDisplayDismissButton);
+        
+        if (offeringId != null) {
+            // TODO: Remove debug log before shipping
+            Log.d(TAG, "Using launchIfNeeded with offering ID");
+            launcher.launchIfNeeded(
+                requiredEntitlementIdentifier,
+                offeringId,
+                new PresentedOfferingContext(offeringId), // TODO: pass PresentedOfferingContext data
```
tonidero: I think this is important for charts to work correctly. Maybe it's ok for now and can be done in a separate PR but we should definitely do this before calling this production ready.

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/RevenueCatUI.java
```diff
@@ -0,0 +1,90 @@
...
+                return;
+            }
+
+            Intent intent = new Intent(activity, PaywallTrampolineActivity.class);
+            
+            if (offeringIdentifier != null && !offeringIdentifier.isEmpty()) {
+                intent.putExtra(EXTRA_OFFERING_ID, offeringIdentifier);
+            }
+            
+            intent.putExtra(EXTRA_SHOULD_DISPLAY_DISMISS_BUTTON, displayCloseButton);
```
tonidero: As mentioned in a previous comment, I think we could pass a single parcelable parameter here, so it's easier to add/modify parameters.

RevenueCatUI/Scripts/IPaywallPresenter.cs
```diff
@@ -11,17 +11,19 @@ internal interface IPaywallPresenter
         /// <summary>
         /// Presents a paywall with the given options.
         /// </summary>
+        /// <param name="gameObjectName">Name of GameObject to receive callbacks</param>
         /// <param name="options">Paywall presentation options</param>
         /// <returns>Result of the paywall presentation</returns>
-        Task<PaywallResult> PresentPaywallAsync(PaywallOptions options);
+        Task<PaywallResult> PresentPaywallAsync(string gameObjectName, PaywallOptions options);
```
tonidero: Hmm so to check... this means that it requires to have a game object to present the paywalls, so we can't call this method directly through code without creating a game object right? I wonder if there is a way we can do this without having to use game objects for people that want to avoid it... We can add wrappers to use the Editor if needed, but I think the "main" API should support not using game objects at all...

Having said this, I might have missed conversations around why this is needed. Happy to discuss it if so!

tonidero (commented):
Looking good! Really glad we'll get this soon 🙌 Just some questions and comments

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallTrampolineActivity.java
vegaro: agree. I was talking to @facumenzella the other day that we can probably prepare a beta without this and custom font support, but we need this to not call it beta. I started adding it but it adds a bit of code so I decided to keep it for a future PR

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/RevenueCatUI.java
vegaro: I like it!

RevenueCatUI/Scripts/IPaywallPresenter.cs
vegaro: oh actually I can remove this after removing the usage of `UnitySendMessage`! Let me clean it up

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallTrampolineActivity.java
JayShortway: Yea that's fair. I'm totally okay going with this for now! 

Down the line we can look at using this pattern in PHC. Maybe we don't _move_  `PaywallTrampolineActivity` to PHC, but we create a separate one. That way we can still avoid the `PaywallFragment`, and don't need to rely on keeping a reference to the listener. That way it would work like this:

```
╭-- Unity-------------------------╮
|           launchPaywall()       |
|                  ||             |
|                  \/             |
|  UnityPaywallTrampolineActivity |
╰---------------------------------╯
                   ||
                   \/ 
╭-- PHC---------------------------╮
|     PaywallTrampolineActivity   |
|                  ||             |
|                  \/             |
|          PaywallActivity        |
╰---------------------------------╯
```

But again, that's something we can worry about later. Let's go with what we have for now!

vegaro: From trampoline to trampoline!

Subtester/Packages/manifest.json
vegaro: yes, this is just a local reference to where the plugin lives

vegaro:
This PR requires https://github.com/RevenueCat/purchases-android/pull/2695/

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallUnityOptions.java
```diff
@@ -0,0 +1,69 @@
+package com.revenuecat.purchasesunity.ui;
+
+import android.os.Parcel;
+import android.os.Parcelable;
+
+import androidx.annotation.Nullable;
+
+public class PaywallUnityOptions implements Parcelable {
```
tonidero, line 8: This would be so good in kotlin with a data class 🥲

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallTrampolineActivity.java
```diff
@@ -0,0 +1,185 @@
...
+
+import androidx.annotation.Nullable;
+import androidx.activity.ComponentActivity;
+
+import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallActivityLauncher;
+import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallResult;
+import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallResultHandler;
+import com.revenuecat.purchases.PresentedOfferingContext;
+
+public class PaywallTrampolineActivity extends ComponentActivity implements PaywallResultHandler {
```
tonidero, line 18: I guess these need to be public to be accessible from C#? Just wondering if we can make them package private or protected to avoid some users accidentally using this code directly (probably NABD, I wouldn't really consider this part of our public API surface...)

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallTrampolineActivity.java
```diff
@@ -0,0 +1,185 @@
...
+        PaywallUnityOptions options = new PaywallUnityOptions(offeringIdentifier, displayCloseButton, requiredEntitlementIdentifier);
+
+        Intent intent = new Intent(activity, PaywallTrampolineActivity.class);
+        intent.putExtra(EXTRA_PAYWALL_OPTIONS, options);
+
+        try {
+            activity.startActivity(intent);
+        } catch (Throwable t) {
+            Log.e(TAG, "Error launching PaywallTrampolineActivity for presentPaywallIfNeeded", t);
+            RevenueCatUI.sendPaywallResult("ERROR");
```
tonidero: Should we extract the different paywall results to constants to avoid repeating them?

RevenueCatUI/Plugins/iOS/RevenueCatUI.m
```diff
@@ -1,31 +1,162 @@
...
+
+static void RCUIPresentPaywallInternal(NSString *offeringIdentifier, BOOL displayCloseButton, RCUIPaywallResultCallback callback) {
+    dispatch_async(dispatch_get_main_queue(), ^{
+        if (@available(iOS 15.0, *)) {
+            __block PaywallProxy *proxy = [[PaywallProxy alloc] init];
+
+            NSMutableDictionary *options = RCUICreateOptionsDictionary(offeringIdentifier, displayCloseButton);
+
+            [proxy presentPaywallWithOptions:options
+                         paywallResultHandler:^(NSString * _Nonnull resultName) {
```
tonidero: Nitpick, but this doesn't seem aligned correctly? Same below in RCUIPresentPaywallIfNeededInternal

RevenueCatUI/Scripts/Platforms/Android/AndroidPaywallPresenter.cs
```diff
@@ -42,8 +41,14 @@ public Task<PaywallResult> PresentPaywallAsync(PaywallOptions options)
             try
             {
                 var offering = options?.OfferingIdentifier;
-                Debug.Log($"[RevenueCatUI][Android] presentPaywall offering='{offering ?? "<null>"}'");
-                _plugin.CallStatic("presentPaywall", offering);
+                var displayCloseButton = options?.DisplayCloseButton ?? false;
```
tonidero: I think the default should be `true`?

RevenueCatUI/Scripts/Platforms/Android/AndroidPaywallPresenter.cs
```diff
@@ -42,8 +41,14 @@ public Task<PaywallResult> PresentPaywallAsync(PaywallOptions options)
...
+                
+                Debug.Log($"[RevenueCatUI][Android] presentPaywall offering='{offering ?? "<null>"}', displayCloseButton={displayCloseButton}");
+                using (var unityPlayer = new AndroidJavaClass("com.unity3d.player.UnityPlayer"))
```
tonidero: Just checking but would this crash if not found? Just thinking we might want to be defensive and react to possible issues if so.

RevenueCatUI/Scripts/Platforms/Android/AndroidPaywallPresenter.cs
```diff
@@ -66,8 +71,13 @@ public Task<PaywallResult> PresentPaywallIfNeededAsync(string requiredEntitlemen
             try
             {
                 var offering = options?.OfferingIdentifier;
-                Debug.Log($"[RevenueCatUI][Android] presentPaywallIfNeeded entitlement='{requiredEntitlementIdentifier}', offering='{offering ?? "<null>"}'");
-                _plugin.CallStatic("presentPaywallIfNeeded", requiredEntitlementIdentifier, offering);
+                var displayCloseButton = options?.DisplayCloseButton ?? false;
+                Debug.Log($"[RevenueCatUI][Android] presentPaywallIfNeeded entitlement='{requiredEntitlementIdentifier}', offering='{offering ?? "<null>"}', displayCloseButton={displayCloseButton}");
+                using (var unityPlayer = new AndroidJavaClass("com.unity3d.player.UnityPlayer"))
```
tonidero: Since the class is repeated, might be good to extract to a constant? We could also extract this block to a method.

RevenueCatUI/Scripts/RevenueCatUI.cs
```diff
@@ -7,22 +7,22 @@ namespace RevenueCat.UI
     /// <summary>
     /// Main interface for RevenueCat UI components.
     /// Provides methods to present paywalls.
+    /// Add this component to a GameObject to enable paywall functionality.
     /// </summary>
-    public static class RevenueCatUI
+    public sealed class RevenueCatUI : MonoBehaviour
```
tonidero: Just to check, does this still need to be a MonoBehavior?

Subtester/Assets/Plugins/Android/settingsTemplate.gradle.DISABLED
```diff
@@ -0,0 +1,26 @@
+pluginManagement {
```
tonidero, line 1: Hmm is this file needed?

Subtester/Assets/Scripts/PurchasesListener.cs
```diff
@@ -196,6 +201,270 @@ private void CreatePurchasePackageForPlacementButtons()
...
+        }
+        else if (offerings?.Current != null)
+        {
+            offeringId = offerings.Current.Identifier;
+        }
+
+        // Create options with specific offering
+        var options = new RevenueCat.UI.PaywallOptions
+        {
+            OfferingIdentifier = offeringId ?? "default",
```
tonidero: Hmm should the default be null instead of `default`?

tonidero (commented):
Looking pretty good! Just some minor questions and comments

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallTrampolineActivity.java
JayShortway: Gymnasts! 💪

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/AndroidManifest.xml
JayShortway: >Like can the `PaywallResultHandler` live longer than the launcher?

Yes, if `launcher` holds a reference to `PaywallResultHandler` (and not the other way around), `PaywallResultHandler` can live longer.

I wonder how it works if we enable _Don't keep activities_, and have `PaywallTrampolineActivity` launch the paywall. My assumption is that the framework handles delivering the paywall result back to `PaywallTrampolineActivity`, even if that was killed in the meantime. We can also check if finishing the `PaywallTrampolineActivity` after launching the paywall has any effect on this.

vegaro: I will double check. That will definitely make this a trampoline, which is beneficial

RevenueCatUI/Scripts/Platforms/Android/AndroidPaywallPresenter.cs
JZDesign: If this is a Unity SDK, what would cause this to be not found that we would want to defend against it instead of let it crash?

JZDesign: I suppose if it's out in the wild, we shouldn't crash and make an app unusable. I get your point, better to be paranoid than to crash a multi-million dollar game

Subtester/Assets/Scripts/PurchasesListener.cs
JZDesign: If this is to render the default paywall (paywall v1) then I don't think it matters. Though it is important to know that there isn't one in macOS at the moment.

JZDesign: The logic to check for the default paywall is if there isn't one found for the given offering id, then fallback to the default.

JZDesign: Oh, this is a test suite isn't it? I would say leaving it as "default" is good here IMO. It's a contextual marker.

Subtester/Assets/Plugins/Android/settingsTemplate.gradle.DISABLED
vegaro: When the custom settings.gradle setting is disabled in the player settings it adds the `DISABLED` to the extension. There's also a setting committed in the Main scene that's set to 0 (disabled), but it looks that Unity changes it to 1 (enabled) and leaves it as 1 even when disabling it in the player settings, it's misbehaving. It looks like having this with `.DISABLED` makes it behave better and doesn't enable the setting automatically.

RevenueCatUI/Scripts/RevenueCatUI.cs
vegaro: I can remove it from this PR. @facumenzella has a PR improving this here https://github.com/RevenueCat/purchases-unity/pull/682

It creates a MonoBehavior and static calls, inspired by #674, and it feels like it's what we should be doing. I can revert the change in this file and merge that other PR later. We wanted to keep it separate to not have too much code in this one, since it was adding up.

RevenueCatUI/Scripts/Platforms/Android/AndroidPaywallPresenter.cs
vegaro: Imho, we should be crashing. If this class is not there, the plugin is not correctly installed.

By looking at the docs https://docs.unity3d.com/Manual/android-high-level-api-code-examples.html

I think I can access UnityPlayer this way `AndroidApplication.unityPlayer`, so I will update it to do that

RevenueCatUI/Scripts/Platforms/Android/AndroidPaywallPresenter.cs
vegaro: good catch!!!

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallTrampolineActivity.java
vegaro: I actually think that we can! I will try that

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallUnityOptions.java
vegaro: I know 😢 . I did try to use Kotlin out of curiosity, and Unity uses Kotlin 1.7.22 by default even in Unity 6.2... That version is not compatible with our plugin and will introduce installation issues to developers

RevenueCatUI/Scripts/Platforms/Android/AndroidPaywallPresenter.cs
tonidero: I agree that not having this available means there is something wrong with the integration... but I'm very reticent to crashing on the SDK, and prefer to just fail the call in a non-crashing way... In this case, I guess I don't have a strong opinion, but still feel that as a rule of thumb, we should avoid crashing.

vegaro: I don't have a strong opinion either, so I can do that. The only thing is that, would you prevent crashing for all `new AndroidJavaClass`? Or only for `new AndroidJavaClass("com.unity3d.player.UnityPlayer")`?

Subtester/Assets/Scripts/PurchasesListener.cs
tonidero: Well, the default offering doesn't necessarily need to have the "default" id in all projects. Some projects might not have an offering with "default" id at all. So in those cases, I think it's better to default to the `current` offering instead. But yeah, this is just for our test app so I guess it's not a big deal.

RevenueCatUI/Scripts/Platforms/Android/AndroidPaywallPresenter.cs
tonidero: Hmm I guess the former, so we avoid crashing in as many situations as possible... but I would say, not a blocker for this PR.

Subtester/Assets/Scripts/PurchasesListener.cs
vegaro: I think I added this to test an offering that is called `default` in our tester app. But I think I should remove it to test the null case, which should show the `current` offering

RevenueCatUI/Scripts/Platforms/Android/AndroidPaywallPresenter.cs
vegaro: Handled it better in [`bcc83a6` (#649)](https://github.com/RevenueCat/purchases-unity/pull/649/commits/bcc83a67dcc0d5aa108cc7c0e265c2fc18d9df23)

RevenueCatUI/Scripts/Platforms/Android/AndroidPaywallPresenter.cs
vegaro: Fixed by using `AndroidApplication` directly

RevenueCatUI/Scripts/RevenueCatUI.cs
vegaro: I reverted the changes in this file in favor of #682 

In any case, we are not shipping this until #682 is merged

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallTrampolineActivity.java
vegaro: Actually, `PaywallTrampolineActivity` needs to be public becasue it's an Activtiy

tonidero: Hmm in Kotlin you can have internal activities... but yeah, that might not work in Java. Let's keep it public then!

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/AndroidManifest.xml
vegaro: btw @JayShortway I did try this (without Don't keep activities):

```
launcher = new PaywallActivityLauncher(this, new PaywallResultHandler() {
            @Override
            public void onActivityResult(PaywallResult o) {
                Log.e("Testing", "PaywallTrampolineActivity onActivityResult: " + o);
            }
        });
```

and finishing the activity right at the end of `onCreate` (after launching the paywall), and the `PaywallResultHandler` doesn't get called, so it looks like the result handler can't survive the Activity

vegaro: Also, I did test "Don't keep activities" and as soon as the paywall is seen, the app restarts. It looks like `UnityPlayerActivity` gets killed and the app is restarted

PR #654, opened 2025-08-29
Bump Subtester Editor version

vegaro:


Subtester/Assets/Plugins/Android/gradleTemplate.properties
```diff
@@ -1,8 +1,5 @@
 org.gradle.jvmargs=-Xmx**JVM_HEAP_SIZE**M -Dfile.encoding=UTF-8 -XX:+UseG1GC -XX:MaxMetaspaceSize=1024m
-org.gradle.parallel=false
-org.gradle.daemon=false
-org.gradle.workers.max=1
-android.enableDexingArtifactTransform=false
+org.gradle.parallel=true
```
vegaro, line 2: This was changed in #655 but I reverted them and tests still pass

.circleci/config.yml
```diff
@@ -96,23 +108,6 @@ commands:
             # rename all the .cs.break files in APITests to .cs
             for file in IntegrationTests/Assets/APITests/*.cs.break ; do mv "$file" "${file%.*}" ; done
 
-  activate-license:
```
vegaro, line 99: this is done using the orb now

.circleci/config.yml
```diff
@@ -203,22 +201,27 @@ jobs:
...
       - run:
           name: Verify APK exists
-          command: test -f Subtester/buildAndroid.apk
+          command: test -f Subtester/Builds/Android/Android.apk
 
       - store_artifacts:
-          path: Subtester/buildAndroid.apk
+          path: Subtester/Builds/Android/Android.apk
 
   build-subtester-ios:
```
vegaro, line 226: This works so I didn't change it to use the orb and it uses our build command. It's basically the same anyways

tonidero: Would it be worth to still use the orb for consistency? (Can be in a separate PR)

IntegrationTests/ProjectSettings/MultiplayerManager.asset
```diff
@@ -0,0 +1,7 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!655991488 &1
+MultiplayerManager:
+  m_ObjectHideFlags: 0
+  m_EnableMultiplayerRoles: 0
+  m_StrippingTypes: {}
```
tonidero, line 7: Is this file needed?

Subtester/Assets/Plugins/Android/baseProjectTemplate.gradle
```diff
@@ -10,7 +10,7 @@ allprojects {
             // See which Gradle version is preinstalled with Unity here https://docs.unity3d.com/Manual/android-gradle-overview.html
             // See official Gradle and Android Gradle Plugin compatibility table here https://developer.android.com/studio/releases/gradle-plugin#updating-gradle
             // To specify a custom Gradle version in Unity, go do "Preferences > External Tools", uncheck "Gradle Installed with Unity (recommended)" and specify a path to a custom Gradle version
-            classpath 'com.android.tools.build:gradle:4.0.1'
+            classpath 'com.android.tools.build:gradle:8.7.2'
```
tonidero: 🎉 

And just to check... would it be worth updating to the latest version? I think `8.12.2` at the moment?

Subtester/ProjectSettings/MultiplayerManager.asset
```diff
@@ -0,0 +1,7 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!655991488 &1
+MultiplayerManager:
+  m_ObjectHideFlags: 0
+  m_EnableMultiplayerRoles: 0
+  m_StrippingTypes: {}
```
tonidero, line 7: Same here about whether this can be removed.

scripts/create-unity-package.sh
```diff
@@ -18,23 +18,111 @@ SYMBOLIC_LINK_PATH="$PROJECT/Assets/RevenueCat"
 MANIFEST_JSON_PATH="$PROJECT/Packages/manifest.json"
 # Here we are adding a symbolic link in the Subtester/Assets project to the RevenueCat scripts so they are exported
 # as part of the .unitypackage
-ln -s "$PWD/RevenueCat" "$PROJECT/Assets/"
+echo "📁 Source RevenueCat folder contents:"
+ls -la "$PWD/RevenueCat/"
+echo ""
+
+if [ ! -z "$CI" ]; then
+    # In CI (Linux), Unity doesn't traverse symlinks properly, so copy the folder
```
tonidero, line 45: Hmm that's weird... not sure why symlinks wouldn't work on Unix 🤔

scripts/create-unity-package.sh
```diff
@@ -61,23 +168,42 @@ fi
...
+        echo "   Package file not found: $PACKAGE"
+    fi
+    # Cleanup RevenueCat folder/symlink
+    rm -rf "$PROJECT/Assets/RevenueCat"
+    exit 1
+fi
 
-rm $SYMBOLIC_LINK_PATH
+# Cleanup RevenueCat folder/symlink
+rm -rf "$PROJECT/Assets/RevenueCat"
```
tonidero, line 266: Not for this PR!! But might be a good idea to move this script to fastlane or something that is easier to use and test... But as I said, this should be done separately.

tonidero (commented):
Still need more review and to test it, but looking great so far!! Excited to get this 🙌 🙌

scripts/create-unity-package.sh
vegaro: I have no idea to be honest but this was causing me trouble. It wouldn't work until I changed the approach to copy files instead of using symlinks

I think it's more of a problem with the new Unity editor in CI (Unix) not with Unix itself, because this was working properly before

Subtester/Assets/Plugins/Android/baseProjectTemplate.gradle
vegaro: I can try. This was the recommended support version according to Unity docs. Gradle comes embedded in Unity so that's why they recommend a particular AGP version

.circleci/config.yml
vegaro: Yeah, definitely worth it. I wanted to keep this PR to the minimum needed and will come in a separate PR

IntegrationTests/ProjectSettings/MultiplayerManager.asset
vegaro: it was automatically created when migrating the project so I figured we would commit it. Maybe it can be ignored?

scripts/create-unity-package.sh
vegaro: Please, it's horrible haha

Subtester/Assets/Plugins/Android/baseProjectTemplate.gradle
vegaro: Alright, as I suspected, bumping this to 8.12.2 won't work because it's incompatible with the version of Gradle that Unity uses. I am checking 8.9.3 which looks like it's compatible with Gradle 8.11

```
A problem occurred evaluating project ':launcher'.
> Failed to apply plugin 'com.android.internal.version-check'.
   > Minimum supported Gradle version is 8.13. Current version is 8.11.
     Try updating the 'distributionUrl' property in /root/project/Subtester/Library/Bee/Android/Prj/Mono2x/Gradle/gradle/wrapper/gradle-wrapper.properties to 'gradle-8.13-bin.zip'.
```

tonidero: It's ok to leave the current version if it works. Was just checking, but probably not worth spending too long there.

Subtester/Assets/Plugins/Android/mainTemplate.gradle
```diff
@@ -13,6 +13,7 @@
 }
 // Android Resolver Repos End
 apply plugin: 'com.android.library'
+apply from: '../shared/keepUnitySymbols.gradle'
```
facumenzella, line 16: Is this file generated on export? 🤔

IntegrationTests/UserSettings/Search.settings
```diff
@@ -0,0 +1 @@
+{}
```
facumenzella: Are these user settings needed? I wonder if we can ignore them

Subtester/Assets/Plugins/Android/mainTemplate.gradle
vegaro: this is generated when enabling the `mainTemplate.gradle` setting in Unity settings. It was also updated when migrating to the new Editor

IntegrationTests/UserSettings/Search.settings
vegaro: we probably can, specially in the `IntegrationTests`

scripts/create-unity-package.sh
```diff
@@ -18,23 +18,111 @@ SYMBOLIC_LINK_PATH="$PROJECT/Assets/RevenueCat"
...
+    
+    # Verify copy was successful
+    if [ -d "$PROJECT/Assets/RevenueCat" ]; then
+        echo "✅ RevenueCat folder copied successfully"
+    else
+        echo "❌ Failed to copy RevenueCat folder!"
+        exit 1
+    fi
+else
+    # Local development (macOS) - use symlink for convenience
```
facumenzella, line 60: Thinking out loud and probably Maybe out of scope, but we (should?) support any OS for local development, right?

vegaro: good point, this script should support windows as well. I am going to rewrite it in ruby or something easier to understand in the close future and will make sure that's supported

scripts/create-unity-package.sh
```diff
@@ -18,23 +18,111 @@ SYMBOLIC_LINK_PATH="$PROJECT/Assets/RevenueCat"
...
+
 # This removes the purchases-unity package dependency from the Subtester project
 # to export it, otherwise it will fail due to duplicated files with the package dependency.
+echo "📄 Original manifest.json:"
+cat $MANIFEST_JSON_PATH
+echo ""
 awk '!/com.revenuecat.purchases-unity/' $MANIFEST_JSON_PATH > temp && mv temp $MANIFEST_JSON_PATH
-FOLDERS_TO_EXPORT=$(cd $PROJECT; find Assets/RevenueCat/* Assets/PlayServicesResolver Assets/ExternalDependencyManager -type d -prune)
+echo "📄 Modified manifest.json:"
+cat $MANIFEST_JSON_PATH
```
facumenzella: What about using a `--verbose` parameter to print in debug mode?

Otherwise it will be too verbose even when you don't care about the logs

PR #655, opened 2025-09-01
Fix false positive Android CI jobs

vegaro:
I have discovered that the Unity CI android checks have been failing with a false positive since as far as I can tell (oldest I can see is 6 months old). Failure is:

```
FAILURE: Build failed with an exception.

* Where:
Build file '/root/project/Subtester/Library/Bee/Android/Prj/Mono2x/Gradle/launcher/build.gradle' line: 1

* What went wrong:
A problem occurred evaluating project ':launcher'.
> Failed to apply plugin [id 'com.android.internal.version-check']
   > Minimum supported Gradle version is 7.0.2. Current version is 6.1.1. If using the gradle wrapper, try editing the distributionUrl in /root/project/Subtester/Library/Bee/Android/Prj/Mono2x/Gradle/gradle/wrapper/gradle-wrapper.properties to gradle-7.0.2-all.zip
```

### DIFF
--- a/IntegrationTests/Assets/Plugins/Android/mainTemplate.gradle
+++ b/IntegrationTests/Assets/Plugins/Android/mainTemplate.gradle
@@ -14,10 +14,6 @@ apply plugin: 'com.android.library'
 
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
-// Android Resolver Dependencies Start
-    implementation 'androidx.annotation:annotation:[1.2.0]' // Assets/RevenueCat/Plugins/Editor/RevenueCatDependencies.xml:5
-    implementation 'com.revenuecat.purchases:purchases-hybrid-common:[14.2.0]' // Assets/RevenueCat/Plugins/Editor/RevenueCatDependencies.xml:4
-// Android Resolver Dependencies End
 **DEPS**}
 
 // Android Resolver Exclusions Start
--- a/IntegrationTests/Assets/Plugins/Android/mainTemplate.gradle
+++ b/IntegrationTests/Assets/Plugins/Android/mainTemplate.gradle
@@ -1,0 +1,66 @@
+// Android Resolver Repos Start
+([rootProject] + (rootProject.subprojects as List)).each { project ->
+    project.repositories {
+        maven {
+            url "https://maven.google.com"
+        }
+        mavenLocal()
+        mavenCentral()
+    }
+}
+// Android Resolver Repos End
+apply plugin: 'com.android.library'
+**APPLY_PLUGINS**
+
+dependencies {
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
+// Android Resolver Dependencies Start
+    implementation 'androidx.annotation:annotation:[1.2.0]' // Assets/RevenueCat/Plugins/Editor/RevenueCatDependencies.xml:5
+    implementation 'com.revenuecat.purchases:purchases-hybrid-common:[14.2.0]' // Assets/RevenueCat/Plugins/Editor/RevenueCatDependencies.xml:4
+// Android Resolver Dependencies End
+**DEPS**}
+
+// Android Resolver Exclusions Start
+android {
+  packagingOptions {
+      exclude ('/lib/arm64-v8a/*' + '*')
+      exclude ('/lib/armeabi/*' + '*')
+      exclude ('/lib/mips/*' + '*')
+      exclude ('/lib/mips64/*' + '*')
+      exclude ('/lib/x86/*' + '*')
+      exclude ('/lib/x86_64/*' + '*')
+  }
+}
+// Android Resolver Exclusions End
+android {
+    compileSdkVersion **APIVERSION**
+    buildToolsVersion '**BUILDTOOLS**'
+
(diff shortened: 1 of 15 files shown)


Subtester/Assets/Plugins/Android/AndroidManifest.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest
     xmlns:android="http://schemas.android.com/apk/res/android"
+    package="com.unity3d.player"
```
tonidero, line 4: Hmm wasn't this deprecated? I think now you need to specify the package in the build.gradle right? The way it was in the `mainTemplate.gradle`... Having said that, when exporting the subtester project, it always adds it for me, and I need to remove if from the AndroidManifest manually 🥴

Subtester/Assets/Plugins/Android/baseProjectTemplate.gradle
```diff
@@ -10,7 +10,7 @@ allprojects {
             // See which Gradle version is preinstalled with Unity here https://docs.unity3d.com/Manual/android-gradle-overview.html
             // See official Gradle and Android Gradle Plugin compatibility table here https://developer.android.com/studio/releases/gradle-plugin#updating-gradle
             // To specify a custom Gradle version in Unity, go do "Preferences > External Tools", uncheck "Gradle Installed with Unity (recommended)" and specify a path to a custom Gradle version
-            classpath 'com.android.tools.build:gradle:8.7.2'
+            classpath 'com.android.tools.build:gradle:4.0.1'
```
tonidero, line 13: 👴

tonidero (commented):
Great catch! Just have a question regarding that change from the mainTemplate to the AndroidManifest... any reason for that change?

Subtester/Assets/Plugins/Android/AndroidManifest.xml
vegaro: I updated the main template on Friday in this PR https://github.com/RevenueCat/purchases-unity/pull/653

The problem is that it looks like the version of Gradle that's embedded in this Unity we are using requires an older version of AGP, which requires this package to be set, that's why I am reverting this change.

In any case, let me put this back into draft and try with AGP 7, since it might be compatible with that Gradle we are using

tonidero: Ohh right, make sense... But yeah... I guess this won't build with Android Studio so it would be great to update the Unity version... but well, if it's too much of a blocker, I think this might be an okish temporal solution.

vegaro: yeah, we are going to have to update manually, as we already had to before https://github.com/RevenueCat/purchases-unity/pull/653. Good news is that it looks like I am managing to build with the new editor in CI, so this shouldn't last for too long

Subtester/Assets/Plugins/Android/mainTemplate.gradle
```diff
@@ -70,12 +68,4 @@ android {
...
 
-// IMPORTANT: If you're not installing through Unity Package Manager, remove the code below.
-// Symlink local sources. When using Unity's "Symlink local sources" checkbox,
-// The link will be wrong when using Unity Package Manager because of a bug in Unity.
-// That's not a problem if it points to a file that doesn't exist, but we need to add the path
-// that does exist here.
-// More info: https://forum.unity.com/threads/native-sources-inside-package-not-found-after-exporting-project-to-android.1355321/
-android.sourceSets.main.java.srcDirs += [
-    '**DIR_UNITYPROJECT**/../RevenueCat/Plugins/Android/'
-]
```
vegaro, line 81: This was causing a duplicated classes issue @tonidero, I removed it and it seems to be working fine

IntegrationTests/Assets/Plugins/Android/mainTemplate.gradle
vegaro, line 1: I had to add this because downloading all the `aar`s was causing duplicated Kotlin classes classes. We might want to recommend using `mainTemplate`s in the docs as well

IntegrationTests/Assets/Plugins/Android/mainTemplate.gradle
```diff
@@ -0,0 +1,62 @@
...
+**APPLY_PLUGINS**
+
+dependencies {
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
+**DEPS**}
+
+// Android Resolver Exclusions Start
+android {
+  packagingOptions {
+      exclude ('/lib/arm64-v8a/*' + '*')
```
tonidero, line 22: So we only include armv7 right? (Probably due to using apretty old version of unity/agp/...

tonidero (approved):
In general I think changes make sense. Hopefully we can get the project updated to latest unity soon though 🙏

IntegrationTests/Assets/Plugins/Android/mainTemplate.gradle
vegaro: Oh this was actually automatically added. The only thing I had to add was the repositories section at the top

PR #671, opened 2025-09-16
Show iOS Paywall 

facumenzella:
## Summary

Implement iOS paywalls in RevenueCatUI with a pure-code callback path by default and an optional UnitySendMessage fallback. Keep core (RevenueCat) UI‑agnostic. Update Subtester to demo both modes.

RevenueCatUI 
      - Present paywall via PurchasesHybridCommonUI PaywallProxy (iOS 15+).
      - Strongly retain PaywallProxy during presentation to guarantee callback delivery.
      - Dual callback paths:
        - Primary: native delegate → completes Task<PaywallResult>. (Make pure-code the default for flexibility, with a GameObject-based fallback only when explicitly enabled)
        - Optional fallback: UnitySendMessage to a receiver (disabled by default).

### DIFF
--- a/RevenueCatUI/Plugins/iOS/RevenueCatUI.m
+++ b/RevenueCatUI/Plugins/iOS/RevenueCatUI.m
@@ -1,37 +1,78 @@
+// RevenueCat UI iOS bridge implemented via PurchasesHybridCommonUI
 #import <Foundation/Foundation.h>
-
-// Minimal native stubs for iOS bridging
+@import PurchasesHybridCommonUI;
+// Import Swift interface for direct calls into PaywallProxy
+#import <PurchasesHybridCommonUI/PurchasesHybridCommonUI-Swift.h>
 
 typedef void (*RCUIPaywallResultCallback)(const char* result);
 typedef void (*RCUICustomerCenterCallback)(void);
 
+static NSString *RCUINormalizeResult(NSString *resultName) {
+    if (resultName == nil) {
+        return @"ERROR";
+    }
+    NSString *lower = [[resultName lowercaseString] stringByReplacingOccurrencesOfString:@"_" withString:@""];
+    if ([lower isEqualToString:@"purchased"]) return @"PURCHASED";
+    if ([lower isEqualToString:@"restored"]) return @"RESTORED";
+    if ([lower isEqualToString:@"cancelled"]) return @"CANCELLED";
+    if ([lower isEqualToString:@"error"]) return @"ERROR";
+    if ([lower isEqualToString:@"notpresented"]) return @"NOT_PRESENTED";
+    return @"ERROR";
+}
+
+static PaywallProxy *s_rcuiPaywallProxy = nil;
+
 void rcui_presentPaywall(const char* offeringIdentifier, bool displayCloseButton, RCUIPaywallResultCallback callback) {
-    NSLog(@"[RevenueCatUI][iOS] presentPaywall(offering=%@, closeButton=%@)",
-          offeringIdentifier ? [NSString stringWithUTF8String:offeringIdentifier] : @"<null>",
-          displayCloseButton ? @"true" : @"false");
-    if (callback) {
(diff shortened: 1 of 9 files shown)


RevenueCatUI/Scripts/RevenueCatUI.cs
```diff
@@ -134,5 +144,28 @@ public static bool IsCustomerCenterSupported()
...
             catch { return false; }
         }
+
+        /// <summary>
+        /// Enables an optional UnitySendMessage fallback for receiving native callbacks.
+        /// Disabled by default. When enabled, a persistent receiver GameObject is created.
+        /// Pure-code (delegate) callbacks remain the primary mechanism.
+        /// </summary>
+        /// <param name="receiverName">Optional receiver GameObject name</param>
+        public static void EnableUnityMessageFallback(string receiverName = null)
```
vegaro: what's this for?

facumenzella: As far as I understand, this is to hook to the messages using `UnitySendMessage`. 
There are two options now:
- Task<PaywallResult> --> pure code way
- UnitySendMessage to a receiver (a GameObject). --> to do this you need to enable it using `EnableUnityMessageFallback`. 

Maybe it's too much?

vegaro: yeah, let's keep it simple for now

Subtester/ProjectSettings/ProjectSettings.asset
```diff
@@ -570,6 +570,9 @@ PlayerSettings:
   - serializedVersion: 3
     m_BuildTarget: Android
     m_Formats: 01000000
+  - serializedVersion: 3
+    m_BuildTarget: iOS
+    m_Formats: 03000000
```
facumenzella:
```suggestion
```

RevenueCatUI/Editor/RevenueCatUIDependencies.xml
vegaro, line 1: In my branch I put this file in `RevenueCatUI/Plugins/Editor/` to match what we do in `RevenueCat`

RevenueCatUI/Editor/RevenueCatUI.Editor.asmdef.meta
vegaro, line 1: I don't think we need this file at all

RevenueCatUI/Plugins/iOS/RevenueCatUI.m
```diff
@@ -1,31 +1,63 @@
+// RevenueCat UI iOS bridge implemented via PurchasesHybridCommonUI
```
vegaro, line 1:
```suggestion
```

RevenueCatUI/Plugins/iOS/RevenueCatUI.m
```diff
@@ -1,31 +1,63 @@
...
+// Import Swift interface for direct calls into PaywallProxy
+#import <PurchasesHybridCommonUI/PurchasesHybridCommonUI-Swift.h>
 
 typedef void (*RCUIPaywallResultCallback)(const char* result);
 
+static NSString *RCUINormalizeResult(NSString *resultName) {
+    if (resultName == nil) {
+        return @"ERROR";
+    }
+    NSString *lower = [[resultName lowercaseString] stringByReplacingOccurrencesOfString:@"_" withString:@""];
```
vegaro, line 13: interesting, should we just check for `not_presented` directly? I was actually a bit confused by the fact that you were checking for `notpresented` without the `_` until I noticed this

RevenueCatUI/Scripts/Platforms/Stub.meta
vegaro, line 1: this file can be removed

Subtester/Assets/Scripts/PurchasesListener.cs
vegaro, line 1: I think there is actually no changes to this file? It can probably be reverted

PR #690, opened 2025-09-30
Update EDM4U in Subtester

vegaro:
We were using a version from 2023

Changelog -> [`update-EDM4U?expand=1`#diff-3e1673b0b8](https://github.com/RevenueCat/purchases-unity/compare/update-EDM4U?expand=1#diff-3e1673b0b81691b891b02d95e567f49245360e2ff0257c988f00ae28a81e5544)


tonidero (approved):
Thank you so much for doing this! 🫶

PR #693, opened 2025-10-01
Change default of `autoSyncPurchases` to `true` on runtime setups

vegaro:
We were wrongly setting `autoSyncPurchases` to `false` on runtime setups. The default is `true` in Android and in `RevenueCat.cs`


facumenzella:
Nice catch,  I think that we should be able to write a test about this 🤔

vegaro:
I agree we do need to add tests, but we don't have any tests currently in place in this repository

tonidero (approved):

PR #700, opened 2025-10-03
Add coverage for API

facumenzella:
## Motivation

Adding some tests to make sure we don't break things while iterating


facumenzella:
I'll open a new one based on the latest PRs

PR #701, opened 2025-10-03
Update Subtester and IntegrationTests to 6000.2.6f2

vegaro:
Required update to fix a security issue https://unity.com/security/sept-2025-01/remediation#faq


vegaro:
ugh, there's some archiving issues in iOS (related to unity classes). I upgraded xcode (we were on 14) and now I see some ruby issues. Will continue next week on this @tonidero @ajpallares

PR #706, opened 2025-10-06
Proper PaywallsBehaviour

vegaro:
We realized the current version wasn't correct since the functions are not available from the Editor. Meaning that if you try to link a Button's `onClick` event to `PresentPaywall`, it's not possible because of the parameters, and becuase it doesn't return `void`.

This PR adds proper configuration options and refactors the functions so they can be worked directly from the Editor.

Here's an example of the component with a handler for the result events attached to it:

Screenshot 2025-10-07 at 10 08 07

And then a Button can be configured to present the paywall:

Screenshot 2025-10-07 at 10 09 08

### DIFF
--- a/Subtester/Assets/Scripts/PurchasesListener.cs
+++ b/Subtester/Assets/Scripts/PurchasesListener.cs
@@ -263,10 +263,7 @@ public class PurchasesListener : Purchases.UpdatedCustomerInfoListener
 
     private System.Collections.IEnumerator PresentPaywallWithOptionsCoroutine()
     {
-        var options = new RevenueCatUI.PaywallOptions
-        {
-            DisplayCloseButton = false
-        };
+        var options = new RevenueCatUI.PaywallOptions(displayCloseButton: false);
 
         var task = RevenueCatUI.PaywallsPresenter.Present(options);
         while (!task.IsCompleted) { yield return null; }


RevenueCatUI/Scripts/PaywallsBehaviour.cs
```diff
@@ -1,32 +1,180 @@
...
+        [Tooltip("Whether to display a close button on the paywall (only for original template paywalls).")]
+        [SerializeField] private bool displayCloseButton = false;
+
+        [Header("Conditional Presentation")]
+        [Tooltip("If set, the paywall will only be presented if the user doesn't have this entitlement.")]
+        [SerializeField] private string requiredEntitlementIdentifier;
+
+        [Header("Events")]
+        [Tooltip("Invoked when the user completes a purchase.")]
+        public UnityEvent OnPurchased = new UnityEvent();
```
vegaro, line 29: These UnityEvents can accept generics, but won't work with a custom class like `PaywallResult`, so I decided to split it into multiple events, that way they can be set in the editor. The alternative is to use a primitive type (string), but I think that limits us more if we ever want to send something in the events:

```c#
[Serializable]
public class StringEvent : UnityEvent<string> { }

[Header("Events")]
public StringEvent OnPaywallResult = new StringEvent();

private void HandleResult(PaywallResult result)
{
    OnPaywallResult?.Invoke(result.Result.ToString());
}
```

vegaro: If we ever need to send something like `UnityEvent<CustomerInfo>`, it's still possible but devs need to hook to it via code (it doesn't show up in the Editor's dropdown):

```c#
[Serializable]
public class CustomerInfoEvent : UnityEvent<CustomerInfo> { }

[Header("Events")]
public CustomerInfoEvent OnCustomerInfoUpdated = new CustomerInfoEvent();
```

```c#
using UnityEngine;

public class CustomerInfoHandler : MonoBehaviour
{
    [SerializeField] private Purchases purchases;

    void Start()
    {
        // Subscribe to the event
        purchases.OnCustomerInfoUpdated.AddListener(HandleCustomerInfo);
    }

    void OnDestroy()
    {
        // Unsubscribe
        purchases.OnCustomerInfoUpdated.RemoveListener(HandleCustomerInfo);
    }

    private void HandleCustomerInfo(Purchases.CustomerInfo customerInfo)
    {
        Debug.Log($"Customer info updated: {customerInfo.ActiveSubscriptions.Count} active subscriptions");
        
        foreach (var entitlement in customerInfo.Entitlements.Active)
        {
            Debug.Log($"Active entitlement: {entitlement.Key}");
        }
    }
}
```

Or we could remove the parameter and inform developers to fetch the CustomerInfo themselves which I believe is the Unity standard since it works in the Editor:

```c#
[Header("Events")]
[Tooltip("Invoked when customer info is updated. Use GetCustomerInfo() to fetch the latest data.")]
public UnityEvent OnCustomerInfoUpdated = new UnityEvent();

private void NotifyCustomerInfoUpdate()
{
    OnCustomerInfoUpdated?.Invoke();
}
```

RevenueCatUI/Scripts/PaywallsBehaviour.cs
```diff
@@ -1,32 +1,180 @@
...
+                    OnCancelled?.Invoke();
+                    break;
+                case PaywallResultType.NotPresented:
+                    OnNotPresented?.Invoke();
+                    break;
+                case PaywallResultType.Error:
+                    OnError?.Invoke();
+                    break;
+            }
         }
```
facumenzella, line 195: Maybe it's worth logging if the callback is not set? 
For example:

` OnError?.Invoke();` if `OnError` is null, this will be lost

vegaro: what do we do if it's not set?

RevenueCatUI/Scripts/PaywallsBehaviour.cs
```diff
@@ -1,32 +1,180 @@
-using System.Threading.Tasks;
+using System;
 using UnityEngine;
+using UnityEngine.Events;
 
 namespace RevenueCatUI
 {
     /// <summary>
-    /// MonoBehaviour helper that forwards to the static PaywallsPresenter API so paywalls can be driven from scenes.
+    /// MonoBehaviour component for presenting RevenueCat paywalls from the Unity Editor.
```
tonidero: Should this be 
```suggestion
    /// MonoBehaviour component for presenting RevenueCat Paywalls from the Unity Editor.
```

RevenueCatUI/Scripts/PaywallsBehaviour.cs
```diff
@@ -1,32 +1,180 @@
...
+        
+        [Tooltip("Whether to display a close button on the paywall (only for original template paywalls).")]
+        [SerializeField] private bool displayCloseButton = false;
+
+        [Header("Conditional Presentation")]
+        [Tooltip("If set, the paywall will only be presented if the user doesn't have this entitlement.")]
+        [SerializeField] private string requiredEntitlementIdentifier;
+
+        [Header("Events")]
+        [Tooltip("Invoked when the user completes a purchase.")]
```
tonidero: Would it make sense to mention this will only be called once the paywall is dismissed? Like
```suggestion
        [Tooltip("Invoked when the user completes a purchase and the paywall is dismissed.")]
```

Same for the other events. I think it could be confusing otherwise.

RevenueCatUI/Scripts/PaywallsBehaviour.cs
```diff
@@ -1,32 +1,180 @@
...
 
         /// <summary>
         /// Presents a paywall only if the user does not have the specified entitlement.
         /// </summary>
-        /// <param name="requiredEntitlementIdentifier">Entitlement identifier to check before presenting.</param>
-        /// <param name="options">Options for presenting the paywall.</param>
-        /// <returns>A <see cref="PaywallResult"/> describing the outcome.</returns>
-        public async Task<PaywallResult> PresentPaywallIfNeeded(string requiredEntitlementIdentifier, PaywallOptions options = null)
+        /// <param name="entitlementIdentifier">Entitlement identifier to check before presenting</param>
+        public async void PresentPaywallIfNeeded(string entitlementIdentifier)
```
tonidero: Hmm I'm just wondering, since whether using `PresentPaywall` or this one depends on the presence of a parameter, I'm wondering if we should unify them into a single method that just uses one or the other depending on that parameter...

tonidero (commented):
Just a couple comments, but looking great!! Excited to get this out!

RevenueCatUI/Scripts/PaywallsBehaviour.cs
vegaro: I think that makes sense yes! I did this in case wanted to call this directly via code. But they can still do it using the setter, or the static version

PR #709, opened 2025-10-07
Add integration tests for PaywallPresenter

facumenzella:
Thank you for contributing to Purchases. Before pressing the "Create Pull Request" button, please provide the following:

  - [ ] A description about what and why you are contributing, even if it's trivial.

  - [ ] The issue number(s) or PR number(s) in the description if you are contributing in response to those.

  - [ ] If applicable, unit tests.


RevenueCatUI/Tests/Runtime/PaywallOptionsAPITests.cs
```diff
@@ -0,0 +1,54 @@
...
+            Assert.IsNotNull(options);
+            Assert.IsNull(options.OfferingIdentifier);
+            Assert.IsFalse(options.DisplayCloseButton);
+        }
+
+        [Test]
+        public void PaywallOptions_OfferingIdentifierConstructor_SetsOfferingIdentifier()
+        {
+            const string offeringId = "test_offering";
+            var options = new PaywallOptions(offeringId);
```
tonidero, line 22: I think this is going to change to accept an offering instead of an ID with #710 right?

RevenueCatUI/Tests/Runtime/PaywallOptionsAPITests.cs
```diff
@@ -0,0 +1,54 @@
+using NUnit.Framework;
+using RevenueCatUI;
+
+namespace RevenueCatUI.Tests
+{
+    public class PaywallOptionsAPITests
```
tonidero, line 6: Hmm it's interesting since we don't have tests for the `RevenueCat` module... are we sure this won't be included in the final bundle? For `RevenueCat`, we have the API tests live in the `IntegrationTests` project instead. Maybe we could follow the same approach by making that project also depend on `RevenueCatUI` and add the tests there?

facumenzella:
I'll do the same as for RevenueCat in another branch 👍

PR #710, opened 2025-10-07
Add `PresentedOfferingContext` support

vegaro:


RevenueCatUI/Scripts/PaywallOptions.cs
```diff
@@ -9,41 +9,42 @@ namespace RevenueCatUI
     public class PaywallOptions
     {
         /// <summary>
-        /// The identifier of the offering to present.
+        /// The offering to present.
         /// If not provided, the current offering will be used.
         /// </summary>
-        public string OfferingIdentifier { get; set; }
+        public Purchases.Offering Offering { get; set; }
```
vegaro, line 15: This is a breaking change on the beta, but will outline it in the changelog

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallTrampolineActivity.java
```diff
@@ -109,6 +114,36 @@ private void launchPaywall(PaywallUnityOptions options) {
         }
     }
 
+    @Nullable
+    private PresentedOfferingContext mapPresentedOfferingContext(@Nullable String jsonString, @Nullable String fallbackOfferingId) {
```
facumenzella, line 118: In this context `jsonString` is `offeringIdentifier` right?

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallTrampolineActivity.java
```diff
@@ -109,6 +114,36 @@ private void launchPaywall(PaywallUnityOptions options) {
...
+            if (targetingMap != null && targetingMap.has("revision") && targetingMap.has("ruleId")) {
+                int revision = targetingMap.optInt("revision");
+                String ruleId = targetingMap.optString("ruleId", null);
+                if (ruleId != null) {
+                    targetingContext = new PresentedOfferingContext.TargetingContext(revision, ruleId);
+                }
+            }
+
+            return new PresentedOfferingContext(offeringIdentifier, placementIdentifier, targetingContext);
+        } catch (JSONException e) {
```
tonidero, line 141: Could be worth logging something though it should never happen hopefully.

RevenueCatUI/Scripts/PaywallOptions.cs
```diff
@@ -9,41 +9,42 @@ namespace RevenueCatUI
...
         /// </summary>
-        /// <param name="offeringIdentifier">The offering identifier to present</param>
-        /// <param name="displayCloseButton">Whether to display a close button</param>
-        public PaywallOptions(string offeringIdentifier, bool displayCloseButton)
+        /// <param name="offering">The offering to present</param>
+        /// <param name="displayCloseButton">Whether to display a close button. Only applicable for original template paywalls, ignored for V2 Paywalls.</param>
+        public PaywallOptions(Purchases.Offering offering, bool displayCloseButton = false)
         {
-            OfferingIdentifier = offeringIdentifier;
+            Offering = offering ?? throw new ArgumentNullException(nameof(offering));
```
tonidero: So I guess they can't pass a null parameter, and instead should use the other constructor? I wonder if we should allow to pass null here for convenience...

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallTrampolineActivity.java
tonidero: `jsonString` may have `offeringIdentifier` but also `placementIdentifer` and/or `targetingContext` I think?

facumenzella: I wonder if we can avoid this? Feels kind of flaky

facumenzella: Mainly because don't control how it's encoded right? The order of encoding / decoding should matter 🤔 
```
var presentedOfferingContextJson = options?.PresentedOfferingContext?.ToJsonString();
```

tonidero: Hmm but this is a json string which gets decoded as a json object so order doesn't really matter I think?

facumenzella: Really? 🥳 
Then nevermind

PR #712, opened 2025-10-08
Add basic interface for Customer Center

facumenzella:
## Motivation
Introduce basic static interface to present Customer Center

> I don't have an Android device to test this 😢

### DIFF
--- a/RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/CustomerCenterTrampolineActivity.java
+++ b/RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/CustomerCenterTrampolineActivity.java
@@ -1,0 +1,69 @@
+package com.revenuecat.purchasesunity.ui;
+
+import android.app.Activity;
+import android.content.Intent;
+import android.os.Bundle;
+import android.util.Log;
+
+import androidx.activity.ComponentActivity;
+import androidx.activity.result.ActivityResultLauncher;
+import androidx.activity.result.contract.ActivityResultContracts;
+import androidx.annotation.Nullable;
+
+import com.revenuecat.purchases.Purchases;
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.CustomerCenterActivity;
+
+public class CustomerCenterTrampolineActivity extends ComponentActivity {
+    private static final String TAG = "PurchasesUnity";
+
+    private static final String RESULT_DISMISSED = "DISMISSED";
+    private static final String RESULT_ERROR = "ERROR";
+
+    private ActivityResultLauncher<Intent> launcher;
+
+    @Override
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        launcher = registerForActivityResult(
+                new ActivityResultContracts.StartActivityForResult(),
+                result -> {
+                    RevenueCatUI.sendCustomerCenterResult(RESULT_DISMISSED);
+                    finish();
+                }
+        );
+
+        if (!Purchases.isConfigured()) {
+            Log.e(TAG, "Purchases is not configured. Cannot launch Customer Center.");
+            RevenueCatUI.sendCustomerCenterResult(RESULT_ERROR);
+            finish();
+            return;
+        }
+
+        try {
+            Intent intent = CustomerCenterActivity.Companion.createIntent$revenuecatui_defaultsRelease(this);
+            launcher.launch(intent);
+        } catch (Throwable t) {
+            Log.e(TAG, "Error launching CustomerCenterActivity", t);
+            RevenueCatUI.sendCustomerCenterResult(RESULT_ERROR);
+            finish();
+        }
+    }
+
+    public static void presentCustomerCenter(Activity activity) {
+        if (activity == null) {
+            Log.e(TAG, "Activity is null; cannot launch Customer Center");
+            RevenueCatUI.sendCustomerCenterResult(RESULT_ERROR);
+            return;
+        }
+
+        Intent intent = new Intent(activity, CustomerCenterTrampolineActivity.class);
+
+        try {
+            activity.startActivity(intent);
+        } catch (Throwable t) {
+            Log.e(TAG, "Error launching CustomerCenterTrampolineActivity", t);
+            RevenueCatUI.sendCustomerCenterResult(RESULT_ERROR);
+        }
+    }
+}
--- a/RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/RevenueCatUI.java
+++ b/RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/RevenueCatUI.java
@@ -5,12 +5,17 @@ import android.util.Log;
 
 public class RevenueCatUI {
     public interface PaywallCallbacks { void onPaywallResult(String result); }
+    public interface CustomerCenterCallbacks { void onCustomerCenterResult(String result); }
 
     private static final String TAG = "RevenueCatUI";
     private static volatile PaywallCallbacks paywallCallbacks;
+    private static volatile CustomerCenterCallbacks customerCenterCallbacks;
 
     public static void registerPaywallCallbacks(PaywallCallbacks cb) { paywallCallbacks = cb; }
     public static void unregisterPaywallCallbacks() { paywallCallbacks = null; }
+
+    public static void registerCustomerCenterCallbacks(CustomerCenterCallbacks cb) { customerCenterCallbacks = cb; }
+    public static void unregisterCustomerCenterCallbacks() { customerCenterCallbacks = null; }
 
     public static void presentPaywall(Activity activity, String offeringIdentifier, boolean displayCloseButton) {
         PaywallTrampolineActivity.presentPaywall(activity, offeringIdentifier, displayCloseButton);
@@ -18,6 +23,10 @@ public class RevenueCatUI {
 
     public static void presentPaywallIfNeeded(Activity activity, String requiredEntitlementIdentifier, String offeringIdentifier, boolean displayCloseButton) {
         PaywallTrampolineActivity.presentPaywallIfNeeded(activity, requiredEntitlementIdentifier, offeringIdentifier, displayCloseButton);
+    }
+
+    public static void presentCustomerCenter(Activity activity) {
+        CustomerCenterTrampolineActivity.presentCustomerCenter(activity);
     }
 
     public static void sendPaywallResult(String result) {
@@ -30,6 +39,19 @@ public class RevenueCatUI {
             }
         } catch (Throwable e) {
             Log.e(TAG, "Error sending paywall result: " + e.getMessage());
+        }
+    }
+
+    public static void sendCustomerCenterResult(String result) {
+        try {
+            CustomerCenterCallbacks cb = customerCenterCallbacks;
+            if (cb != null) {
+                cb.onCustomerCenterResult(result);
+            } else {
+                Log.w(TAG, "No callback registered to receive customer center result: " + result);
+            }
+        } catch (Throwable e) {
+            Log.e(TAG, "Error sending customer center result: " + e.getMessage());
         }
     }
 }
--- a/RevenueCatUI/Scripts/CustomerCenterPresenter.cs
+++ b/RevenueCatUI/Scripts/CustomerCenterPresenter.cs
@@ -1,0 +1,34 @@
+using System;
+using System.Threading.Tasks;
+using UnityEngine;
+using RevenueCatUI.Internal;
+
+namespace RevenueCatUI
+{
+    /// <summary>
+    /// Main interface for presenting the RevenueCat Customer Center experience.
+    /// </summary>
+    public static class CustomerCenterPresenter
+    {
+        /// <summary>
+        /// Presents the Customer Center UI modally.
+        /// </summary>
+        /// <returns>A task describing the outcome of the presentation.</returns>
+        public static async Task<CustomerCenterResult> Present()
+        {
+            try
+            {
+                Debug.Log("[RevenueCatUI] Presenting Customer Center...");
+                var presenter = CustomerCenterPlatformPresenter.Instance;
+                var result = await presenter.PresentAsync();
+                Debug.Log($"[RevenueCatUI] Customer Center finished with result: {result.Result}");
+                return result;
+            }
+            catch (Exception e)
+            {
+                Debug.LogError($"[RevenueCatUI] Error presenting Customer Center: {e.Message}");
+                return CustomerCenterResult.Error;
+            }
+        }
+    }
+}
--- a/RevenueCatUI/Scripts/Platforms/Android/AndroidCustomerCenterPresenter.cs
+++ b/RevenueCatUI/Scripts/Platforms/Android/AndroidCustomerCenterPresenter.cs
@@ -1,0 +1,114 @@
+#if UNITY_ANDROID && !UNITY_EDITOR
+using System;
+using System.Threading.Tasks;
+using RevenueCatUI.Internal;
+using UnityEngine;
+using UnityEngine.Android;
+
+namespace RevenueCatUI.Platforms
+{
+    internal class AndroidCustomerCenterPresenter : ICustomerCenterPresenter
+    {
+        private readonly AndroidJavaClass _plugin;
+        private readonly CallbacksProxy _callbacks;
+        private TaskCompletionSource<CustomerCenterResult> _current;
+
+        public AndroidCustomerCenterPresenter()
+        {
+            try
+            {
+                _plugin = new AndroidJavaClass("com.revenuecat.purchasesunity.ui.RevenueCatUI");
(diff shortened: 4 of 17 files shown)


RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/CustomerCenterTrampolineActivity.java
```diff
@@ -0,0 +1,69 @@
...
+    private static final String RESULT_ERROR = "ERROR";
+
+    private ActivityResultLauncher<Intent> launcher;
+
+    @Override
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        launcher = registerForActivityResult(
+                new ActivityResultContracts.StartActivityForResult(),
```
vegaro: This should be `new ShowCustomerCenter(),` which is the contract for opening the Customer Center

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/CustomerCenterTrampolineActivity.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        try {
+            Intent intent = CustomerCenterActivity.Companion.createIntent$revenuecatui_defaultsRelease(this);
+            launcher.launch(intent);
+        } catch (Throwable t) {
+            Log.e(TAG, "Error launching CustomerCenterActivity", t);
+            RevenueCatUI.sendCustomerCenterResult(RESULT_ERROR);
+            finish();
+        }
```
vegaro, line 57: After the change to use `ShowCustomerCenter()`, this should be:

```suggestion
        try {
            launcher.launch(Unit.INSTANCE);
        } catch (Throwable t) {
            Log.e(TAG, "Error launching Customer Center", t);
            RevenueCatUI.sendCustomerCenterResult(RESULT_ERROR);
            finish();
        }
```

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/CustomerCenterTrampolineActivity.java
```diff
@@ -0,0 +1,69 @@
...
+import com.revenuecat.purchases.Purchases;
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.CustomerCenterActivity;
+
+public class CustomerCenterTrampolineActivity extends ComponentActivity {
+    private static final String TAG = "PurchasesUnity";
+
+    private static final String RESULT_DISMISSED = "DISMISSED";
+    private static final String RESULT_ERROR = "ERROR";
+
+    private ActivityResultLauncher<Intent> launcher;
```
vegaro:
```suggestion
    private ActivityResultLauncher<Unit> launcher;
```

vegaro:
Regarding the Material2 stuff. It happens in all hybrids, and we are going to fix it in purchases-android

RevenueCatUI/Scripts/CustomerCenterPresenter.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// Presents the Customer Center UI modally.
+        /// </summary>
+        /// <returns>A task describing the outcome of the presentation.</returns>
+        public static async Task<CustomerCenterResult> Present()
+        {
+            try
+            {
+                Debug.Log("[RevenueCatUI] Presenting Customer Center...");
+                var presenter = CustomerCenterPlatformPresenter.Instance;
+                var result = await presenter.PresentAsync();
```
vegaro: I think there is an API design issue here. Unlike Paywalls, we don't want a single result for the Customer Center, and we want a callback/listener system, right?

With the current implementation, there can only be a single result. But in Flutter and React Native there's a set of callbacks

facumenzella: I've introduced `[CustomerCenterCallbacks.cs‎](https://github.com/RevenueCat/purchases-unity/pull/712#diff-445bf4395869a3aa95b205ffec7d6537c1001a0b70e370b7077f448cc1234d7a)`

This would look more or less like:

```swift
  async void ShowCustomerCenter()
  {
      var callbacks = new CustomerCenterCallbacks
      {
          OnDismissed = reason => Debug.Log($"Customer Center dismissed: {reason}"),
          OnError = error => Debug.LogError($"Customer Center error: {error.Code} ({error.Message})"),
          OnEvent = payload =>
          {
              if (payload.Type == CustomerCenterEventType.LinkTapped)
              {
                  Debug.Log($"Help link tapped: {payload.LinkUrl}");
              }
          }
      };

      var start = await CustomerCenterPresenter.Present(callbacks);

      if (start.Status != CustomerCenterStartStatus.Launched)
      {
          Debug.LogWarning($"Customer Center did not launch: {start.Status}");
      }
  }
```

RevenueCatUI/Scripts/CustomerCenterPresenter.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// Main interface for presenting the RevenueCat Customer Center experience.
+    /// </summary>
+    public static class CustomerCenterPresenter
+    {
+        /// <summary>
+        /// Presents the Customer Center UI modally.
+        /// </summary>
+        /// <param name="callbacks">Placeholder for future callback support.</param>
+        /// <returns>A task that completes when the presentation ends.</returns>
+        public static async Task Present(CustomerCenterCallbacks callbacks = null)
```
vegaro, line 33: I would remove the `CustomerCenterCallbacks` completely from here. Just in case we decide to go with the static listener route, so we don't have to break the API

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/RevenueCatUI.java
```diff
@@ -32,4 +41,17 @@ public static void sendPaywallResult(String result) {
             Log.e(TAG, "Error sending paywall result: " + e.getMessage());
         }
     }
-}
\ No newline at end of file
+
+    public static void sendCustomerCenterResult(String result) {
```
vegaro: Is this ever called at all? Do we need it?

facumenzella: Nope, this is a leftover of the result api

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/RevenueCatUI.java
```diff
@@ -3,9 +3,24 @@
...
 import android.util.Log;
 
+import androidx.annotation.Nullable;
+
 public class RevenueCatUI {
     public interface PaywallCallbacks { void onPaywallResult(String result); }
-    public interface CustomerCenterCallbacks { void onCustomerCenterResult(String result); }
+    
+    public interface CustomerCenterCallbacks {
+        void onCustomerCenterDismissed();
```
facumenzella, line 12: do we want to keep dismissed?

tonidero: Hmm considering the API returns a Future, I think the best would be not to add it. And just return the promise when the customer center is dismissed.

vegaro: Note that this is not public. I am using it to communicate to the C# layer that the Future should be completed. 

This will call `AndroidCustomerCenterPresenter.onCustomerCenterDismissed`, which calls 
`OnCustomerCenterDismissed` and finishes the Task (`_current.TrySetResult(true)`)

vegaro: let me clarify, it's public in Java, but not in C#

RevenueCatUI/Scripts/CustomerCenterPresenter.cs
facumenzella: as agreed in https://github.com/RevenueCat/purchases-flutter/pull/1476, we'll keep it right?

RevenueCatUI/Scripts/Platforms/Android/AndroidCustomerCenterPresenter.cs
```diff
@@ -0,0 +1,295 @@
...
+            _storedCallbacks?.OnCustomActionSelected?.Invoke(
+                new CustomActionSelectedEventArgs(actionId, purchaseIdentifier)
+            );
+        }
+
+        private class CallbacksProxy : AndroidJavaProxy
+        {
+            private readonly AndroidCustomerCenterPresenter _owner;
+
+            public CallbacksProxy(AndroidCustomerCenterPresenter owner) : base("com.revenuecat.purchasesunity.ui.RevenueCatUI$CustomerCenterCallbacks")
```
tonidero: It's weird to have a public constructor and methods for a private class 😅. Should it be internal?

RevenueCatUI/Scripts/Platforms/iOS/IOSCustomerCenterPresenter.cs
```diff
@@ -0,0 +1,243 @@
...
+                            case "change_plans":
+                                optionEnum = CustomerCenterManagementOption.ChangePlans;
+                                break;
+                            default:
+                                optionEnum = CustomerCenterManagementOption.Unknown;
+                                break;
+                        }
+                        
+                        s_storedCallbacks?.OnManagementOptionSelected?.Invoke(
+                            new ManagementOptionSelectedEventArgs(optionEnum, url));
```
tonidero: Total Nitpick, but would this be better?
```suggestion
                            new ManagementOptionSelectedEventArgs(optionEnum, url)
                        );
```

I see this is like this in a bunch of places, so we would need to change it in all of them, but not a blocker.

RevenueCatUI/Scripts/CustomerCenterCallbacks.cs
```diff
@@ -0,0 +1,157 @@
...
+        RefundRequest,
+        ChangePlans,
+        Unknown
+    }
+
+    public sealed class FeedbackSurveyCompletedEventArgs
+    {
+        public string FeedbackSurveyOptionId { get; }
+
+        public FeedbackSurveyCompletedEventArgs(string feedbackSurveyOptionId)
```
tonidero: I wonder if we could make these constructors internal...

RevenueCatUI/Scripts/CustomerCenterCallbacks.cs
```diff
@@ -0,0 +1,157 @@
...
+namespace RevenueCatUI
+{
+    public enum RefundRequestStatus
+    {
+        Success,
+        UserCancelled,
+        Error
+    }
+
+    public enum CustomerCenterManagementOption
```
tonidero: Should we use strings as constants instead of enums? (Since we want to avoid using enums moving forward). I guess this does mean it would be quite different from the other SDKs though... So I'm ok with this in this case I guess.

RevenueCatUI/Scripts/CustomerCenterPresenter.cs
```diff
@@ -0,0 +1,48 @@
...
+        ///     OnFeedbackSurveyCompleted = (args) => Debug.Log($"Survey completed: {args.FeedbackSurveyOptionId}")
+        /// };
+        /// await CustomerCenterPresenter.Present(callbacks);
+        /// </code>
+        /// </example>
+        public static async Task Present(CustomerCenterCallbacks callbacks = null)
+        {
+            try
+            {
+                Debug.Log("[RevenueCatUI] Presenting Customer Center...");
```
tonidero: I'm slightly worried about all these logs possibly being annoying... Ideally we would have a system to join them with the existing `logLevel` and `logHandler` properties, so it's more configurable... But I guess we can leave it and remove if it's noisy for customers.

Subtester/Assets/Scripts/PurchasesListener.cs
```diff
@@ -1,1282 +1,1345 @@
...
+        CreateButton("Purchase Product For WinBack Testing", PurchaseProductForWinBackTesting);
+        CreateButton("Fetch & Redeem WinBack for Product", FetchAndRedeemWinBackForProduct);
+        CreateButton("Purchase Package For WinBack Testing", PurchasePackageForWinBackTesting);
+        CreateButton("Fetch & Redeem WinBack for Package", FetchAndRedeemWinBackForPackage);
+        CreateButton("Get Storefront", GetStorefront);
+        CreateButton("Present Paywall", PresentPaywallResult);
+        CreateButton("Present Paywall with Options", PresentPaywallWithOptions);
+        CreateButton("Present Paywall for Offering", PresentPaywallForOffering);
+        CreateButton("Present Paywall If Needed", PresentPaywallIfNeeded);
+        CreateButton("Present Customer Center", PresentCustomerCenter);
```
tonidero: For reviewing this file, I had to hide whitespace.

Subtester/Assets/Scripts/PurchasesListener.cs
```diff
@@ -1,1282 +1,1345 @@
...
+                Debug.Log($"Subtester: OnManagementOptionSelected - Option: {args.Option}{urlInfo}");
+            },
+            OnCustomActionSelected = (args) =>
+            {
+                string purchaseInfo = args.PurchaseIdentifier != null ? $", Purchase: {args.PurchaseIdentifier}" : "";
+                Debug.Log($"Subtester: OnCustomActionSelected - Action: {args.ActionId}{purchaseInfo}");
+            }
+        };
+
+        var task = RevenueCatUI.CustomerCenterPresenter.Present(callbacks);
```
tonidero: I'm wondering if we should offer an API that doesn't use coroutines... But I think we can start with this and iterate if needed 👍

Subtester/Assets/Scripts/PurchasesListener.cs
```diff
@@ -1,1282 +1,1345 @@
...
+        }
+    }
+
+    private System.Collections.IEnumerator PresentCustomerCenterCoroutine()
+    {
+        var callbacks = new RevenueCatUI.CustomerCenterCallbacks
+        {
+            OnFeedbackSurveyCompleted = (args) =>
+            {
+                Debug.Log($"Subtester: OnFeedbackSurveyCompleted - Option ID: {args.FeedbackSurveyOptionId}");
```
tonidero: Should we also update the infoLabel with these events? Mostly asking since it could be useful when testing on a device

tonidero (approved):
Looking great!! Amazing job with this guys! Left some comments but no blockers.

RevenueCatUI/Scripts/Platforms/Android/AndroidCustomerCenterPresenter.cs
vegaro: oh good catch. It's weird you can define public constructors for public classes in C#. I can make the constructor internal I think since it's called from within this file

vegaro: the methods need to stay public so the JNI bridge can call them. The class can stay private though. I think that's the most encapsulated we can go

Subtester/Assets/Scripts/PurchasesListener.cs
vegaro: These callbacks execute on a background thread (web have documentation for that in CustomerCenterCallbacks), which prevents direct Unity UI manipulation without additional main thread dispatching. I did start making those updates to dispatch to the main thread, but the Customer Center presents as a modal overlay that blocks the underlying Unity scene, so any `infoLabel` updates would only become visible after the Customer Center dismisses, the updates will be queued up while the Customer Center is presented and become effective when the scene is resumed.

This is why I kept them as logs, to simplify things a bit.

tonidero: Ah true... makes sense then, thanks for the context!

Subtester/Assets/Scripts/PurchasesListener.cs
vegaro: Yes, there's actually Async/Await and Coroutines in C#. Let's see how this is received by the community. They might actually ask for Coroutines instead 😆

RevenueCatUI/Scripts/CustomerCenterPresenter.cs
vegaro: I agree. I removed some and changed some to LogWarning instead. Let's see if they are noisy yeah.

RevenueCatUI/Scripts/CustomerCenterCallbacks.cs
vegaro: better to make the change now. I will make it. Making them constants makes sense. Devs will only do string comparison so no need to make an enum.

PR #726, opened 2025-10-17
Introduce CustomerCenter MonoBehaviour

facumenzella:
## Motivation

Compliment static interface with MonoBehaviour interface, following the same pattern as we have for Paywalls


vegaro:
closing for now. We can revisit this if needed

PR #729, opened 2025-10-22
Enable releasing RevenueCatUI

vegaro:
This will make sure the RevenueCatUI Unity package starts getting added to the releases


JZDesign (approved):
elmo fire

PR #916, opened 2026-05-01
Add SPM support for iOS dependencies

rickvdl:
## Description

CocoaPods is being deprecated. EDM4U 1.2.187 (Jan 2026) added Swift Package Manager support for iOS dependencies, so this PR offers SPM as a path forward.

- Declare `PurchasesHybridCommon` and `PurchasesHybridCommonUI` as `<remoteSwiftPackage>` entries in both `RevenueCatDependencies.xml` and `RevenueCatUIDependencies.xml`, alongside the existing `<iosPods>` entries.
- Use `replacesPod` so EDM4U strips the matching pods at resolve time when SPM is honored.
- Drop a redundant `#import <PurchasesHybridCommonUI/PurchasesHybridCommonUI-Swift.h>` from `RevenueCatUI.m` — the `@import PurchasesHybridCommonUI;` already exposes the same `@objc` symbols via Clang modules, and the angle-bracket header doesn't exist under SPM. Matches the pattern used in `purchases-flutter`.
- Bump EDM4U in the Subtester to 1.2.187 and toggle `SwiftPackageManagerEnabled` on by default so contributors test SPM going forward.
- Update the `update_hybrid_common` fastlane lane to also bump the SPM version, so future PHC bumps don't silently leave SPM pinned at an old version.
- Split `build-integration-tests-ios` and `archive-ios` into parameterized jobs that run in parallel — one variant per dependency manager — so both paths stay covered in CI.
- Remove unused `com.unity.ads` package from the IntegrationTests project — it was a template leftover with no code references, and its presence was preventing a clean pure-SPM build, because the package is only available through CocoaPods at the moment.

## Customer impact

**Existing CocoaPods customers are unaffected.** EDM4U < 1.2.187 silently ignores the new `<remoteSwiftPackage>` element and falls through to the existing `<iosPod>` entries, so their builds continue working unchanged.

**SPM is enabled by default in EDM4U 1.2.187+.** Customers who update to EDM4U 1.2.187+ will have SPM active automatically — the `replacesPod` mechanism removes the matching pods from the generated Podfile and adds `purchases-hybrid-common.git` as an SPM dependency to the Xcode project. Customers who prefer CocoaPods can disable SPM via the "Swift Package Manager Enabled" toggle in iOS Resolver settings, and the existing `<iosPod>` entries will be used as normal.

## Testing

Verified locally in the Subtester on a real iOS device, both with and without the SPM toggle enabled — CocoaPods and SPM both build and run successfully.

CI runs both paths in parallel and both archived cleanly. Verified the dependency manager actually used by inspecting the xcodebuild logs:

- [`archive-ios-cocoapods`](https://app.circleci.com/pipelines/github/RevenueCat/purchases-unity/4031/workflows/d17e1c95-e701-42cc-a5e2-35c334e058c0/jobs/20350) — pulled PHC + PHC UI via CocoaPods.
- [`archive-ios-spm`](https://app.circleci.com/pipelines/github/RevenueCat/purchases-unity/4031/workflows/d17e1c95-e701-42cc-a5e2-35c334e058c0/jobs/20355) — resolved via Swift Package Manager:
  ```
  Resolved source packages:
    PurchasesHybridCommon: https://github.com/RevenueCat/purchases-hybrid-common.git @ 18.2.0
    RevenueCat: https://github.com/RevenueCat/purchases-ios-spm @ 5.69.0
  resolved source packages: PurchasesHybridCommon, RevenueCat
  ```

## Required status checks (follow-up)

To avoid breaking branch protection, the SPM variants reuse the original job names (`build-integration-tests-ios` and `archive-ios`) — only the CocoaPods variants get explicit `-cocoapods` suffixes. The existing required checks keep working unchanged.

As a follow-up after merge, we should add `build-integration-tests-ios-cocoapods` and `archive-ios-cocoapods` to the required status checks list so the legacy path stays gated until we drop CocoaPods support entirely.


.circleci/config.yml
```diff
@@ -360,46 +379,54 @@ jobs:
...
-          working_directory: IntegrationTests/buildiOS
+          working_directory: IntegrationTests/buildiOS-<< parameters.variant >>
           command: bundle exec pod install --repo-update --verbose
 
       - save_cache:
-          key: 1-pods-{{ checksum "RevenueCat/Plugins/Editor/RevenueCatDependencies.xml" }}
+          key: 1-pods-<< parameters.variant >>-{{ checksum "RevenueCat/Plugins/Editor/RevenueCatDependencies.xml" }}
           paths:
-            - IntegrationTests/buildiOS/Pods
+            - IntegrationTests/buildiOS-<< parameters.variant >>/Pods
```
ajpallares: Is there any point in restoring/saving the `/Pods` cache when `variant` is `spm`?

.circleci/config.yml
```diff
@@ -360,46 +379,54 @@ jobs:
...
 
       - restore_cache:
-          key: 1-pods-{{ checksum "RevenueCat/Plugins/Editor/RevenueCatDependencies.xml" }}
+          key: 1-pods-<< parameters.variant >>-{{ checksum "RevenueCat/Plugins/Editor/RevenueCatDependencies.xml" }}
 
       - run:
           name: Install Pods
-          working_directory: IntegrationTests/buildiOS
+          working_directory: IntegrationTests/buildiOS-<< parameters.variant >>
           command: bundle exec pod install --repo-update --verbose
```
ajpallares: Should this step be skipped when `variant` is `spm`?

ajpallares (commented):
Thanks for adding this!!

I have some comments. But my biggest concern is that the PR description mentions that using SPM is opt-in, while EDM4U 1.2.187 does this in its [IOSResolver.cs](https://github.com/googlesamples/unity-jar-resolver/blob/v1.2.187/source/IOSResolver/src/IOSResolver.cs#L1193-L1199)

```
public static bool SwiftPackageManagerEnabled {
    get { return settings.GetBool(PREFERENCE_SWIFT_PACKAGE_MANAGER_ENABLED,
                                  defaultValue: true); }
    ...
}
```

which seems to mean that SPM is now enabled by default? 

In fact, on CI, `build-integration-tests-ios` is actually disabling SPM, so it seems to be opt-out.

rickvdl:
> Thanks for adding this!!
> 
> I have some comments. But my biggest concern is that the PR description mentions that using SPM is opt-in, while EDM4U 1.2.187 does this in its [IOSResolver.cs](https://github.com/googlesamples/unity-jar-resolver/blob/v1.2.187/source/IOSResolver/src/IOSResolver.cs#L1193-L1199)
> 
> ```
> public static bool SwiftPackageManagerEnabled {
>     get { return settings.GetBool(PREFERENCE_SWIFT_PACKAGE_MANAGER_ENABLED,
>                                   defaultValue: true); }
>     ...
> }
> ```
> 
> which seems to mean that SPM is now enabled by default?
> 
> In fact, on CI, `build-integration-tests-ios` is actually disabling SPM, so it seems to be opt-out.

That's a very good point! And it seems you're right, it is indeed the default after updating. Updated the PR description to clarify.

.circleci/config.yml
rickvdl: Good idea, updated

.circleci/config.yml
rickvdl: No Indeed we can optimize that, updated :)

ajpallares (commented):
Thanks for iterating on this! I think it looks great! However, the CI is failing [here](https://app.circleci.com/pipelines/github/RevenueCat/purchases-unity/4089/workflows/1e07034b-3b79-4564-bdd2-5ff5a7a794dc/jobs/20810). We should fix it before merging

rickvdl:
> > Thanks for adding this!!
> > I have some comments. But my biggest concern is that the PR description mentions that using SPM is opt-in, while EDM4U 1.2.187 does this in its [IOSResolver.cs](https://github.com/googlesamples/unity-jar-resolver/blob/v1.2.187/source/IOSResolver/src/IOSResolver.cs#L1193-L1199)
> > ```
> > public static bool SwiftPackageManagerEnabled {
> >     get { return settings.GetBool(PREFERENCE_SWIFT_PACKAGE_MANAGER_ENABLED,
> >                                   defaultValue: true); }
> >     ...
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > which seems to mean that SPM is now enabled by default?
> > In fact, on CI, `build-integration-tests-ios` is actually disabling SPM, so it seems to be opt-out.
> 
> That's a very good point! And it seems you're right, it is indeed the default after updating. Updated the PR description to clarify.

@ajpallares 

> Thanks for iterating on this! I think it looks great! However, the CI is failing [here](https://app.circleci.com/pipelines/github/RevenueCat/purchases-unity/4089/workflows/1e07034b-3b79-4564-bdd2-5ff5a7a794dc/jobs/20810). We should fix it before merging

This actually turned out to be because of the `pod install` call was still required (Unity Ads is still only available through CP). AFAIK we aren't using Unity Ads at all, so I've removed the dependency, which then only requires the `.xcodeproj` file.

IntegrationTests/Packages/manifest.json
```diff
@@ -2,8 +2,7 @@
   "dependencies": {
     "com.unity.2d.sprite": "1.0.0",
     "com.unity.2d.tilemap": "1.0.0",
-    "com.unity.ads": "4.12.0",
-    "com.unity.ai.navigation": "2.0.9",
+"com.unity.ai.navigation": "2.0.9",
```
ajpallares: This indentation seems off

IntegrationTests/Packages/packages-lock.json
```diff
@@ -15,16 +15,7 @@
...
-      "version": "4.12.0",
-      "depth": 0,
-      "source": "registry",
-      "dependencies": {
-        "com.unity.ugui": "1.0.0"
-      },
-      "url": "https://packages.unity.com"
-    },
-    "com.unity.ai.navigation": {
+"com.unity.ai.navigation": {
```
ajpallares: Same comment about indentation

RevenueCatUI/Plugins/Editor/RevenueCatUIDependencies.xml
```diff
@@ -5,6 +5,10 @@
     <androidPackage spec="androidx.activity:activity:1.8.2" />
   </androidPackages>
 
+  <remoteSwiftPackage url="https://github.com/RevenueCat/purchases-hybrid-common.git" version="18.2.0">
```
ajpallares: Oh, this version should be updated to match the latest
```suggestion
  <remoteSwiftPackage url="https://github.com/RevenueCat/purchases-hybrid-common.git" version="18.8.0">
```

RevenueCat/Plugins/Editor/RevenueCatDependencies.xml
```diff
@@ -4,6 +4,9 @@
         <androidPackage spec="com.revenuecat.purchases:purchases-hybrid-common:[18.8.0]" />
         <androidPackage spec="androidx.annotation:annotation:[1.2.0]" />
     </androidPackages>
+    <remoteSwiftPackage url="https://github.com/RevenueCat/purchases-hybrid-common.git" version="18.2.0">
```
ajpallares:
```suggestion
    <remoteSwiftPackage url="https://github.com/RevenueCat/purchases-hybrid-common.git" version="18.8.0">
```

ajpallares (commented):
Some new comments. The drift of PHC between SPM and CocoaPods is important, otherwise the automatic update with `files_to_update_phc_version` probably won't work.

Also, the PR description could probably add a quick mention that we're dropping the `com.unity.ads` Unity package from the integration tests

rickvdl:
> Some new comments. The drift of PHC between SPM and CocoaPods is important, otherwise the automatic update with `files_to_update_phc_version` probably won't work.
> 
> Also, the PR description could probably add a quick mention that we're dropping the `com.unity.ads` Unity package from the integration tests

Good catch! All updated in https://github.com/RevenueCat/purchases-unity/pull/916/commits/3039900945c674cd0c2d48d0278a0050a482a47a

ajpallares (approved):
Great job! I think this is ready to ship 🚀

PR #926, opened 2026-05-12
Fix Android paywall ignoring bottom safe area in fullscreen Unity host

tonidero:
## Summary

**Bug**
The Unity Android paywall captured system-bar insets from the host activity's decor view. When the host hides system bars (e.g. Unity with `unity.launch-fullscreen=True`), the activity's `WindowInsetsCompat.getInsets(navigationBars())` returns zero. The dialog opens its own window which still shows the nav bar, so the Compose paywall received a zero bottom inset and drew behind the visible nav bar.

**Fix**
Read insets from the dialog window itself via an `OnApplyWindowInsetsListener` on the dialog's content container. The dialog window reports the bars that are actually visible to the dialog, regardless of what the host activity is doing.

**Flag sequencing**
`FLAG_LAYOUT_NO_LIMITS` is now applied by the listener after the first inset dispatch (previously it was set immediately after `dialog.show()`). This avoids mutating window flags during the very first inset traversal.

**Cache (defensive)**
The listener caches the last non-zero typed insets and re-injects them if a dispatch ever arrives with zero values. This is defensive: on API 30+, typed insets (`Type.statusBars()` and `Type.navigationBars()`) are descriptive and keep being dispatched with real values across rotation, fold/unfold, and multi-window resize regardless of `FLAG_LAYOUT_NO_LIMITS`. The flag only zeroes the legacy `getSystemWindowInsets()` surface, which this code does not read. The cache covers older APIs and edge dispatches that might transiently carry zero typed values.

## Screenshots
| Version | Before | After |
| --------| -------|------|
| 36 | image | image |
| 34 | image | image |

Note how in Android 34 the dialog doesn't show full screen anymore. This is because we don't apply the `FLAG_LAYOUT_NO_LIMITS` flag unless we discover any insets. I think it makes more sense since the default is to not go full screen in Android < 35, as opposed to before, where we were going full screen always. Lmk if you think otherwise.


RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallViewPresenter.java
```diff
@@ -371,70 +372,89 @@ public void onRestoreCompleted(@NonNull CustomerInfo customerInfo) {
     // region Edge-to-edge
 
     /**
-     * Wraps the PaywallView in a container that provides corrected window insets.
+     * Wraps the PaywallView in a container whose insets listener captures the dialog
+     * window's real system-bar insets on first dispatch, then triggers FLAG_LAYOUT_NO_LIMITS.
+     * Once the flag is applied, subsequent dispatches arrive with zero insets — the listener
+     * re-injects the cached values so the PaywallView's Compose content stays padded.
```
ajpallares: Is it possible for this to change at runtime (e.g. if the device rotates) and have the cached values become invalid?

ajpallares (commented):
I think this looks good, but I will leave approval to someone with more knowledge in the Android layout system. I have one question though

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallViewPresenter.java
tonidero: Good question! and I checked and this actually reapplies the insets, I believe because on rotation, the entire activity and dialog is recreated. So this shouldn't be a problem.

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallViewPresenter.java
```diff
@@ -371,70 +372,101 @@ public void onRestoreCompleted(@NonNull CustomerInfo customerInfo) {
...
-        window.setStatusBarColor(Color.TRANSPARENT);
-        window.setNavigationBarColor(Color.TRANSPARENT);
-        window.addFlags(WindowManager.LayoutParams.FLAG_LAYOUT_NO_LIMITS);
-    }
-
-    private static int getSystemBarHeight(Activity activity, String resourceName) {
-        int resourceId = activity.getResources().getIdentifier(resourceName, "dimen", "android");
-        if (resourceId > 0) {
-            return activity.getResources().getDimensionPixelSize(resourceId);
+        if (Build.VERSION.SDK_INT < 35) { // Build.VERSION_CODES.VANILLA_ICE_CREAM
```
JayShortway: I'd pick either the named constant (`VANILLA_ICE_CREAM`) or the integer, not both 😛 For instance:

```suggestion
        if (Build.VERSION.SDK_INT < 35) {
```

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallViewPresenter.java
JayShortway: Is that always the case in Unity? What if the Activity has `android:configChanges=["orientation"]`? (Although I believe that is [ignored](https://developer.android.com/about/versions/16/behavior-changes-16#ignore-orientation) when targeting Android 16.)

JayShortway (approved):

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallViewPresenter.java
tonidero: Right... It failed to find the version with the older versions used in the repo, so I switched to using the integer and left the name of the version as a comment for convenience... But it's not really needed, can remove it.

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/src/main/java/com/revenuecat/purchasesunity/ui/PaywallViewPresenter.java
tonidero: Hmm not sure I understand... if the activity has that, it would lock the orientation (in versions < 16 as you said), so the insets wouldn't change from the initial ones? Probably missing something 😅

PR #927, opened 2026-05-12
feat: ad tracking API

peterporfy:
## Summary

Adds ad impression and revenue tracking support to the Unity SDK.

## Design notes

- All types in `namespace RevenueCat` — no extra `using` needed beyond `using RevenueCat;`
- `purchases.AdTracker` returns a `RevenueCat.AdTracker` instance (initialized in `Purchases.Start()`)
- Value types nested in `AdTracker` (`AdTracker.Format`, `AdTracker.MediatorName`, `AdTracker.Precision`) to avoid naming collisions with ad SDK types like `GoogleMobileAds.Api.AdFormat`. This is different compared to the rest of the SDKs as C# `using` directives do wildcard namespace injection and that results in an immediate `AdFormat` clash between RC and Google. Since we are nesting under `AdTracker` I dropped the `Ad` prefix from the Format, Mediator etc types.
- Data is passed as JSON strings across the C#/native boundary, consistent with existing SDK patterns
- Optional string fields (`networkName`, `placement`) are omitted from JSON when null: SimpleJSON's `Escape()` crashes on null strings, so null checks are required at serialization time
- `mediatorErrorCode` (int?) follows the same pattern: key omitted when null

## Other notes

- I am using simple `remarks` (https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags#remarks) to signal the API is still experimental. I didn't find anything stronger. I considered https://learn.microsoft.com/en-us/dotnet/api/system.obsoleteattribute?view=netframework-4.8.1&viewFallbackFrom=net-10.0 - but it is semantically incorrect, and https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-12.0/experimental-attribute - but it is not supported in older cs.

## Tests

- added API surface tests
- extended the sample app locally and manually tested builds both with android and ios target


RevenueCat/Scripts/AdDisplayedData.cs
```diff
@@ -0,0 +1,43 @@
+using RevenueCat.SimpleJSON;
+
+namespace RevenueCat
+{
+    /// <remarks>Experimental: this API is unstable and may change in a future release.</remarks>
+    public class AdDisplayedData
```
rickvdl, line 6: Should we (at least) add some (de)serialization tests for these data objects to ensure the proper JSON is produced? Especially since the note about SimpleJSON's crash and the workaround :)

RevenueCat/Scripts/AdDisplayedData.cs
```diff
@@ -0,0 +1,43 @@
...
+            var obj = new JSONObject();
+            obj["mediatorName"] = MediatorName.Value;
+            obj["adFormat"] = AdFormat.Value;
+            obj["adUnitId"] = AdUnitId;
+            obj["impressionId"] = ImpressionId;
+            if (NetworkName != null) obj["networkName"] = NetworkName;
+            if (Placement != null) obj["placement"] = Placement;
+            return obj.ToString();
+        }
+    }
```
rickvdl, line 50: Should we implement custom equality and hashable conformances as well? Currently comparing two data objects will just compare their references. Not sure if this is something we might use in the future (the developer might), but this could cause unintended results.

I believe implementing these should work
```
public override bool Equals(object obj) => obj is MediatorName other && other.Value == Value;
public override int GetHashCode() => Value?.GetHashCode() ?? 0;
public static bool operator ==(MediatorName a, MediatorName b) => Equals(a, b);
public static bool operator !=(MediatorName a, MediatorName b) => !Equals(a, b);
```

rickvdl (commented):
Nice work! Two comments, and like the Flutter PR I think we should update the label to use `pr:feat` instead of `pr:other` :)

RevenueCat/Scripts/AdDisplayedData.cs
peterporfy: I don't think it is very likely that someone would do it - but good catch anyway, given this is the behavior everywhere else! Fixed

RevenueCat/Scripts/AdDisplayedData.cs
peterporfy: I can do that - it is in fact missing for existing checks too. The problem is that we don't have a test suite like that currently. APITests are just testing if the api is compiling, but we have no functionality tested as I see. Should I introduce it or I am missing something here?

rickvdl: Right, understand. In that case it may be out of scope for this PR. Would be nice to add this at some point (in general) though :)

RevenueCat/Scripts/PurchasesWrapperiOS.cs
```diff
@@ -537,5 +537,30 @@ public void TrackCustomPaywallImpression(Purchases.CustomPaywallImpressionParams
     {
         _RCTrackCustomPaywallImpression(parameters.PaywallId, parameters.OfferingId);
     }
+
+    [DllImport("__Internal")]
+    private static extern void _RCTrackAdDisplayed(string dataJson);
+    public void TrackAdDisplayed(RevenueCat.AdDisplayedData data) =>
```
rickvdl: Nit: In some places we're using the imported types after `using RevenueCat;` and in some places we're using the fully qualified name like `RevenueCat.AdDisplayedData`. Would be nice to keep this consistent

RevenueCat/Scripts/AdDisplayedData.cs
```diff
@@ -0,0 +1,43 @@
...
+            string placement = null)
+        {
+            MediatorName = mediatorName;
+            AdFormat = adFormat;
+            AdUnitId = adUnitId;
+            ImpressionId = impressionId;
+            NetworkName = networkName;
+            Placement = placement;
+        }
+
```
rickvdl, line 30: Nit: these data types do not implement `ToString()` which makes it a bit harder to debug / print.

rickvdl (approved):
Two small comments, but nothing blocking, looks good to me!

PR #943, opened 2026-06-04
Expose OriginalJson and Signature on StoreTransaction

tonidero:
Some developers have asked for access to the purchase `OriginalJson` and `Signature` properties so they can verify purchases on their own backend without using `PurchasesAreCompletedBy.MY_APP`.

This exposes both on the `StoreTransaction` model. They are optional: only populated on Android (`Signature` is Google Play only), and `null` on iOS and some stores.

Depends on the corresponding purchases-hybrid-common change: RevenueCat/purchases-hybrid-common#1668


ajpallares (approved):
Thanks for adding this!

PR #945, opened 2026-06-05
fix(android): remove unused RevenueCatUI ProGuard config

rickvdl:
## Motivation

`RevenueCatUI.androidlib/build.gradle` referenced `consumer-proguard.pro`, but the package did not ship that file. Newer Android Gradle Plugin versions fail when merging ProGuard files.

## Fix

Removed the unused `consumerProguardFiles` reference instead of adding a new consumer ProGuard file.

RevenueCatUI does not appear to need custom consumer ProGuard rules for Unity classes, and adding them would unnecessarily force downstream apps to keep `com.unity3d.player` classes unobfuscated. Also removed the old local `proguard-android-optimize.txt` file, which contained the same unnecessary rule.

Fixed the `proguard-rules.pro` filename typo while I was at it.

## Testing

Tested locally using a small validation script by Claude

### DIFF
--- a/RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/build.gradle
+++ b/RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/build.gradle
@@ -47,7 +47,7 @@ android {
     buildTypes {
         release {
             minifyEnabled false
-            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules. pro'
+            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
         }
     }
 }
--- a/RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/build.gradle
+++ b/RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/build.gradle
@@ -35,7 +35,6 @@ android {
     def unityLib = project(':unityLibrary').extensions.getByName('android')
 
     defaultConfig {
-        consumerProguardFiles "consumer-proguard.pro"
         minSdkVersion unityLib.defaultConfig.minSdkVersion.mApiLevel
         targetSdkVersion unityLib.defaultConfig.targetSdkVersion.mApiLevel
     }
--- a/RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/consumer-proguard.pro
+++ b/RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/consumer-proguard.pro
@@ -1,0 +1,1 @@
+-keep class com.unity3d.player.** { *; }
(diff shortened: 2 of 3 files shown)


RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/consumer-proguard.pro
```diff
@@ -0,0 +1 @@
+-keep class com.unity3d.player.** { *; }
```
tonidero: Hmm so this is a copy of the `proguard-android-optimize.txt` right? I'm not sure if it's needed... and considering that this has worked for folks for a long time, I'm thinking it might be worth to just remove the `consumerProguardFiles "consumer-proguard.pro"` line above instead? If it worked without this before, we avoid obfuscating more without need. Lmk if you think otherwise @JayShortway @AlvaroBrey

tonidero (commented):
I think I would prefer to just remove the reference if it's worked without it so far. But happy to hear other thoughts!

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/build.gradle
```diff
@@ -47,7 +47,7 @@ android {
     buildTypes {
         release {
             minifyEnabled false
-            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules. pro'
+            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
```
AlvaroBrey, line 49: This line is doing nothing (`minifyEnabled false`), but still good to fix the typo

RevenueCatUI/Plugins/Android/RevenueCatUI.androidlib/consumer-proguard.pro
AlvaroBrey: The only reason we would need this is if we were (in our own code) reading classes or attributes from `com.unity3d.player`, which is [not the case as far as I can see](https://github.com/search?q=repo%3ARevenueCat%2Fpurchases-unity+path%3A%2F%5ERevenueCatUI%5C%2FPlugins%5C%2FAndroid%5C%2FRevenueCatUI%5C.androidlib%5C%2F%2F+com.unity3d&type=code)

The rule also wasn't even being applied to our lib build anyway since we have `minifyEnabled false` so no minification was happening anyway.

In fact, by adding this rule we're needlessly preventing consumers of our lib from obfuscating any unity classes, which wasn't the case before. Unity should (hopefully) be providing its own consumer rules.

So, +1 on just removing `consumerProguardFiles`, and removing `proguard-android-optimize.txt` too.

JayShortway: +2 :D 

If we have no reason for proguard rules, we can just remove the config option from `build.gradle`.

rickvdl: Thanks for the input all! I've removed the file :)

JayShortway (approved):

PR #949, opened 2026-06-10
refactor: Add keep rules for java types

tonidero:
We're working to narrow down our consumer rules in https://github.com/RevenueCat/purchases-android/pull/3557. However, this makes it not find our code anymore in Unity when minifying since nothing is directly referencing the Java types from Unity. This adds some Keep annotations that from my tests avoid the issue. This does mean we will need to remember to mark all java types as `@Keep` but it was easier than trying to add proguard rules to be used by our customers when importing our plugins.

Note this won't do anything until the purchases-android changes actually get merged


ajpallares (approved):
Thanks!

> This does mean we will need to remember to mark all java types as @Keep

Is there any way we can enforce this in our PRs?

tonidero:
Added a check with a bash script in a separate PR: https://github.com/RevenueCat/purchases-unity/pull/953

PR #953, opened 2026-06-11
ci: Enforce @Keep on public Android bridge types

tonidero:
Adds a CI check (`scripts/check-android-keep-annotations.sh`, wired into the `deploy-check` workflow) that fails the build if any **public** Java/Kotlin type under the Unity Android source roots is missing `@Keep`. We're tightening our proguard rules soon and this will prepare this SDK so we don't forget to keep that.


scripts/check-android-keep-annotations.sh
```diff
@@ -0,0 +1,85 @@
...
+
+set -euo pipefail
+
+cd "$(dirname "$0")/.."
+
+# Directories that hold the Android bridge compiled into consumer apps.
+ROOTS=(
+    "RevenueCat/Plugins/Android"
+    "RevenueCatUI/Plugins/Android"
+)
```
ajpallares: Maybe a small thing, but what if we ever change these folders? I guess this would be checking an empty directory. Should we perhaps assert that each root exists or fail if any root has 0 files?

scripts/check-android-keep-annotations.sh
```diff
@@ -0,0 +1,85 @@
...
+        {
+            t = trim($0)
+            is_comment = (t ~ /^(\/\/|\/\*|\*)/)
+
+            # Remember @Keep seen in the current contiguous annotation/comment block.
+            if (!is_comment && t ~ /^@Keep([ \t(]|$)/) seen_keep = 1
+
+            is_type = (!is_comment && t ~ decl_re)
+            if (lang == "kotlin")
+                is_public = (t !~ /(^|[ \t])(private|internal|protected)([ \t])/)
```
ajpallares:
```
class Foo private constructor()
```

Would this be a false negative?

scripts/check-android-keep-annotations.sh
```diff
@@ -0,0 +1,85 @@
...
+                    found = 1
+                }
+                seen_keep = 0
+                next
+            }
+            # A type decl line (even non-public) still ends the annotation block.
+            if (is_type) { seen_keep = 0; next }
+
+            # Annotations, comments and blank lines preserve the flag; other code clears it.
+            if (!(t == "" || is_comment || t ~ /^@/)) seen_keep = 0
```
ajpallares:
```
@Keep
@SuppressWarnings(
    "unchecked"
)
public class MultiLineAnnotation { }
```

Would this be a false positive?

scripts/check-android-keep-annotations.sh
```diff
@@ -0,0 +1,85 @@
...
+    awk -v file="$file" -v lang="$lang" '
+        function trim(s) { gsub(/^[ \t]+|[ \t]+$/, "", s); return s }
+        BEGIN {
+            kw = (lang == "kotlin") ? "class|interface|object" : "class|interface|enum"
+            decl_re = "(^|[^A-Za-z0-9_])(" kw ")[ \t]+[A-Za-z_]"
+            name_re = ".*(" kw ")[ \t]+"
+        }
+        {
+            t = trim($0)
+            is_comment = (t ~ /^(\/\/|\/\*|\*)/)
```
ajpallares:
```java
/*
public class InsideBlockComment {
*/
```

Would this be a false positive?

ajpallares (commented):
Thanks for adding this! Cursor gave me some edge cases that may be worth considering

scripts/check-android-keep-annotations.sh
tonidero: Good call. The script now asserts every root in `ROOTS` exists and that the combined `find` returns at least one file, failing with a clear message otherwise, so a moved/renamed folder can't silently turn the check into a no-op. (126e297)

scripts/check-android-keep-annotations.sh
tonidero: Yes, that was a false negative. Visibility is now decided from the modifiers *before* the type keyword only (via `match`/`RSTART`), so a `private constructor` no longer masks a public class. `class Foo private constructor()` is now correctly flagged. (126e297)

scripts/check-android-keep-annotations.sh
tonidero: Yes, false positive. The scanner now tracks parenthesis depth: once an annotation opens unbalanced `(`, following lines are treated as annotation continuation and don't reset the `@Keep` flag. Multi-line annotations no longer cause a false alarm. (126e297)

scripts/check-android-keep-annotations.sh
tonidero: Yes, false positive. Lines are now run through a comment stripper that drops `//` tails and `/* */` spans (including multi-line, via a persistent in-block flag), so commented-out declarations are ignored. As a bonus it also fixes the inverse miss of `/* x */ public class Foo` on one line. (126e297)

scripts/check-android-keep-annotations.sh
```diff
@@ -0,0 +1,145 @@
...
+                    printf "%s:%d: public type \047%s\047 is missing @Keep\n", file, NR, name
+                    found = 1
+                }
+                seen_keep = 0
+                next
+            }
+
+            # Annotation line (no declaration on it).
+            if (substr(t, 1, 1) == "@") {
+                if (t ~ /^@Keep([ \t(]|$)/) seen_keep = 1
```
ajpallares: Could this be a new false positive?

```
@SuppressWarnings("unchecked") @Keep
public class KeepAfterOtherAnnotation { }
```

ajpallares (approved):
I think there's a new false positive, but other than that, I believe it's good to ship. 

I wonder if we should have a set of fixtures with these edge cases to test that it works as expected, but I think it's overkill so probably we don't need to do this.

Thanks for adding this and iterating on this!

scripts/check-android-keep-annotations.sh
tonidero: Fixed in https://github.com/RevenueCat/purchases-unity/pull/953/commits/15eb43f23a34047334745a032bdbe400732291e3

tonidero:
I kinda thought the same... as in, we're testing the tests 😅 But with all the extra complexity, I'm thinking it might be worth it 😅

PR #959, opened 2026-06-15
Add presentedOfferingContext support to custom paywall impression events

rickvdl:
Based on PHC PR https://github.com/RevenueCat/purchases-hybrid-common/pull/1665.

## API

We will now send the `presentedOfferingContext` from the passed `Offering` along with the event. This PR adds the new API variant that takes an `Offering` instead of a plain string `offeringId`. The latter is now deprecated in favor of this new API variant. We will derive the `presentedOfferingContext` from this `Offering` and send it along with the request.

```
Purchases.TrackCustomPaywallImpression(
    Purchases.CustomPaywallImpressionParams.FromOffering(offering, paywallId: "")
);
```

---

> [!NOTE]
> **Low Risk**
> Analytics/API extension with backward-compatible obsolete overloads; no purchase or auth changes.
> 
> **Overview**
> Custom paywall impression tracking now accepts an **`Offering`** on `CustomPaywallImpressionParams` so analytics can include **placement and targeting** via `presentedOfferingContext` (taken from the first package’s context). The **`offeringId`-only constructor is marked obsolete**; callers should pass an `Offering` instead.
> 
> **Android and iOS** bridges forward a third argument—serialized `presentedOfferingContext`—into hybrid common, with **null-stripping** on parsed JSON before the event is sent. Integration API tests cover both legacy and offering-based call paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5cf862bcc91b17ebc0de185897fb37eb38738f18. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
+++ b/RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
@@ -580,8 +580,8 @@ signedDiscountTimestamp:(NSString *)signedDiscountTimestamp {
 }
 
 - (void)trackCustomPaywallImpression:(nullable NSString *)paywallId
-                           offeringId:(nullable NSString *)offeringId
-             presentedOfferingContext:(nullable NSString *)presentedOfferingContextJson {
+                            offeringId:(nullable NSString *)offeringId
+              presentedOfferingContext:(nullable NSString *)presentedOfferingContextJson {
     NSMutableDictionary *data = [NSMutableDictionary dictionary];
     if (paywallId) {
         data[@"paywallId"] = paywallId;
--- a/RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
+++ b/RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
@@ -10,6 +10,12 @@
 @import PurchasesHybridCommon;
 @import RevenueCat;
 
+@interface NSObject (NSNullMapping)
+
+- (id)mappingNSNullToNil;
+
+@end
+
 static NSString *const RECEIVE_STOREFRONT = @"_receiveStorefront";
 static NSString *const RECEIVE_PRODUCTS = @"_receiveProducts";
 static NSString *const RECEIVE_CUSTOMER_INFO = @"_receiveCustomerInfo";
@@ -56,38 +62,6 @@ char *makeStringCopy(NSString *nstring) {
     strcpy(res, string);
 
     return res;
-}
-
-static id valueByRemovingNSNull(id value) {
-    if (!value || value == (id)[NSNull null]) {
-        return nil;
-    }
-
-    if ([value isKindOfClass:NSDictionary.class]) {
-        NSMutableDictionary *filteredDictionary = [NSMutableDictionary dictionary];
-        NSDictionary *dictionary = (NSDictionary *) value;
-        for (id key in dictionary) {
-            id filteredValue = valueByRemovingNSNull(dictionary[key]);
-            if (filteredValue) {
-                filteredDictionary[key] = filteredValue;
-            }
-        }
-        return [NSDictionary dictionaryWithDictionary:filteredDictionary];
-    }
-
-    if ([value isKindOfClass:NSArray.class]) {
-        NSMutableArray *filteredArray = [NSMutableArray array];
-        NSArray *array = (NSArray *) value;
-        for (id item in array) {
-            id filteredItem = valueByRemovingNSNull(item);
-            if (filteredItem) {
-                [filteredArray addObject:filteredItem];
-            }
-        }
-        return [NSArray arrayWithArray:filteredArray];
-    }
-
-    return value;
 }
 
 #pragma mark RCPurchases Wrapper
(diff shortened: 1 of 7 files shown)


RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
```diff
@@ -573,15 +579,30 @@ - (void)purchasePackageWithWinBackOffer:(NSString *)packageIdentifier presentedO
     }];
 }
 
-- (void)trackCustomPaywallImpression:(nullable NSString *)paywallId offeringId:(nullable NSString *)offeringId {
+- (void)trackCustomPaywallImpression:(nullable NSString *)paywallId
+                           offeringId:(nullable NSString *)offeringId
```
tonidero: Nitpick but seems the indentation on these 2 lines do not match the first one?

RevenueCat/Scripts/CustomPaywallImpressionParams.cs
```diff
@@ -12,20 +14,79 @@ public class CustomPaywallImpressionParams
...
+
+        /// <summary>
+        /// Creates parameters for a custom paywall impression from an Offering.
+        /// </summary>
+        /// <param name="offering">An optional offering associated with the custom paywall. When provided,
+        /// the SDK derives the offering identifier and presented offering context from this offering's first
+        /// available package.</param>
+        /// <param name="paywallId">An optional identifier for the custom paywall being shown.</param>
+        /// <param name="offeringId">Deprecated. This value is ignored when an Offering is provided.</param>
+        public static CustomPaywallImpressionParams FromOffering(Offering offering, string paywallId = null, string offeringId = null)
```
tonidero: We shouldn't be receiving the `offeringId` as a parameter in this constructor I think?

RevenueCat/Scripts/CustomPaywallImpressionParams.cs
```diff
@@ -12,20 +14,79 @@ public class CustomPaywallImpressionParams
 
         /// <summary>
         /// An optional identifier for the offering associated with the custom paywall.
-        /// If not provided, the SDK will use the current offering identifier from the cache.
+        /// If neither this nor an Offering is provided, no offering data is sent and the
+        /// native SDK handles its default fallback behavior.
+        /// Deprecated: use FromOffering instead so the SDK can derive placement and targeting context.
         /// </summary>
+        [Obsolete("Use FromOffering instead.", false)]
```
tonidero: Hmm so this is the property, not the constructor... I wonder if it might be ok to not deprecate this one? Would avoid some pragma checks later and I don't think it's needed... We might want to consider adding the `Offering` itself and in that case we could consider deprecating this one, but then, we shouldn't add the `PresentedOfferingContext` either.

tonidero (commented):
Just a couple small things but almost there!

RevenueCat/Scripts/CustomPaywallImpressionParams.cs
rickvdl: Good catch, same thing as I did for KMP. Will remove that one

RevenueCat/Scripts/CustomPaywallImpressionParams.cs
rickvdl: Ah that's a good idea point, removed the deprecation there and cleaned up the pragma marks. Still kept the presentedOfferingContext there as that kept things simple.

RevenueCat/Scripts/CustomPaywallImpressionParams.cs
```diff
@@ -12,20 +14,74 @@ public class CustomPaywallImpressionParams
 
         /// <summary>
         /// An optional identifier for the offering associated with the custom paywall.
-        /// If not provided, the SDK will use the current offering identifier from the cache.
+        /// If neither this nor an Offering is provided, no offering data is sent and the
+        /// native SDK handles its default fallback behavior.
```
tonidero: TBH, I think this comment is a bit out of place in the property and should be part of the constructor comments...

RevenueCat/Scripts/CustomPaywallImpressionParams.cs
```diff
@@ -12,20 +14,74 @@ public class CustomPaywallImpressionParams
...
         /// Creates parameters for a custom paywall impression.
         /// </summary>
         /// <param name="paywallId">An optional identifier for the custom paywall being shown.</param>
         /// <param name="offeringId">An optional identifier for the offering associated with the custom paywall.
-        /// If not provided, the SDK will use the current offering identifier from the cache.</param>
+        /// If neither this nor an Offering is provided, no offering data is sent and the native SDK
+        /// handles its default fallback behavior.
+        /// Deprecated: use FromOffering instead so the SDK can derive placement and targeting context.</param>
+        [Obsolete("Use FromOffering instead.", false)]
         public CustomPaywallImpressionParams(string paywallId = null, string offeringId = null)
```
tonidero: Hmm just curious how does C# disambiguate between this constructor and the no parameters one? Since it could be both I think... In any case, if we don't get a deprecated warning when no parameters are passed, I think it's ok.

RevenueCat/Scripts/CustomPaywallImpressionParams.cs
```diff
@@ -12,20 +14,74 @@ public class CustomPaywallImpressionParams
...
+        }
+
+        /// <summary>
+        /// Creates parameters for a custom paywall impression from an Offering.
+        /// </summary>
+        /// <param name="offering">An optional offering associated with the custom paywall. When provided,
+        /// the SDK derives the offering identifier and presented offering context from this offering's first
+        /// available package.</param>
+        /// <param name="paywallId">An optional identifier for the custom paywall being shown.</param>
+        public static CustomPaywallImpressionParams FromOffering(Offering offering, string paywallId = null)
```
tonidero: It's a bit weird this is using a factory method instead of using a normal constructor as the others... What's the reason to make it a factory method?

tonidero (approved):
Left some other comments... but I think this can be shipped :)

RevenueCat/Scripts/CustomPaywallImpressionParams.cs
rickvdl: Agree, will move these relevant parts to the constructors and just mention the fact that it's deprecated here

RevenueCat/Scripts/CustomPaywallImpressionParams.cs
rickvdl: It prefers the no-argument one actually :) So it should not trigger the deprecation warning

RevenueCat/Scripts/CustomPaywallImpressionParams.cs
rickvdl: You're right... it was suggested by Codex as more C#-like, but I agree this is inconsistent now. Will remove it in favor of the plain params init.

PR #960, opened 2026-06-15
Sub tester: Allow HTTPS request debugging and fix offerings logging crash

rickvdl:
Two small subtester changes:

- Adds a `network_security_config` that allows for intercepting HTTPS request in order to debug them using a proxy
- Updated offerings response logging from logging the full response to just logging a summary per offering, this avoids crashes when a large offerings response is received

---

> [!NOTE]
> **Medium Risk**
> Trusting user CAs weakens TLS pinning assumptions and is appropriate for a tester build but would be risky if shipped to production; offerings logging change is low-risk tester UI only.
> 
> **Overview**
> **Android (Subtester)** wires the app to a **network security config** that trusts both system and **user-installed** certificate authorities, so HTTPS traffic can be inspected with a debugging proxy. This is delivered as a small Unity Android library plus `android:networkSecurityConfig` on the main application manifest.
> 
> **Offerings screen** stops logging the full `offerings.ToString()` after fetch/sync and instead uses a new **`LogOfferingsSummary`** helper that prints current offering id, count, and per-offering package/context lines—avoiding crashes when offerings payloads are very large.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f3d3d4f6f69dfcdd1df4c07a7ed7e03b1646fbc2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Subtester/Assets/Plugins/Android/network-security-config.androidlib/build.gradle
+++ b/Subtester/Assets/Plugins/Android/network-security-config.androidlib/build.gradle
@@ -1,0 +1,20 @@
+apply plugin: 'com.android.library'
+
+android {
+    namespace 'com.revenuecat.paywall_tester.networksecurity'
+    compileSdk getProperty("unity.compileSdkVersion") as int
+
+    sourceSets {
+        main {
+            manifest.srcFile 'src/main/AndroidManifest.xml'
+            res.srcDirs = ['src/main/res']
+        }
+    }
+
+    def unityLib = project(':unityLibrary').extensions.getByName('android')
+
+    defaultConfig {
+        minSdkVersion unityLib.defaultConfig.minSdkVersion.mApiLevel
+        targetSdkVersion unityLib.defaultConfig.targetSdkVersion.mApiLevel
+    }
+}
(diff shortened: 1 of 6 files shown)


Subtester/Assets/Plugins/Android/network-security-config.androidlib/build.gradle
```diff
@@ -0,0 +1,20 @@
+apply plugin: 'com.android.library'
+
+android {
+    namespace 'com.revenuecat.paywall_tester.networksecurity'
```
tonidero, line 4: It's so difficult to configure these things for Unity :( Need to add a whole separate module just for this 😅

tonidero (approved):
TBH, not sure if there is a simpler way to add this in Unity than adding a new module... But well, I think it makes sense for subtester :) Thank you for the improvements!!

Subtester/Assets/Plugins/Android/network-security-config.androidlib/build.gradle
rickvdl: Yeah 😓 Unfortunately the path of just adding a resource to `Assets/Plugins/Android/res`, but unfortunately that is explicitly unsupported now since Unity 6 :(

PR #980, opened 2026-07-07
(breaking) Change SubscriptionOption.Price.AmountMicros from int to long to avoid overflow

AlvaroBrey:
## Motivation

On Android, `SubscriptionOption.Price.AmountMicros` comes from Google Play Billing as micro-units of the local currency. High nominal prices (for example KRW subscriptions) can exceed `int.MaxValue`. The field was an `int`, so SimpleJSON parsed via `AsInt` / `(int)AsDouble` and saturated to `2147483647`, which produced incorrect prices in custom paywalls.

## Description

Change `SubscriptionOption.Price.AmountMicros` from `int` to `long` so large values are preserved. Update the compile-only API surface test in `SubscriptionOptionAPITests`.

**Breaking change:** Any code that explicitly treats `AmountMicros` as `int` must use `long` instead.

Until this ships, apps can display `PricingPhase.Price.Formatted` rather than deriving display text from `AmountMicros`.

---

> [!NOTE]
> **Medium Risk**
> Public API breaking change (`int` → `long`) affects any paywall or pricing code that reads `AmountMicros`; behavior fix is localized to price parsing with no auth or payment-flow changes.
> 
> **Overview**
> Fixes incorrect Android subscription prices when Google Play **amount micros** exceed `int.MaxValue` (e.g. high nominal currencies like KRW), which previously clamped to `2147483647` and broke custom paywalls that derive display from micro-units.
> 
> **`SubscriptionOption.Price.AmountMicros`** is changed from **`int`** to **`long`** in `SubscriptionOption.cs`. The integration API compile test in **`SubscriptionOptionAPITests`** is updated to match.
> 
> **Breaking:** consumer code that types **`AmountMicros`** as **`int`** must use **`long`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b92d87714b5bc3e0fa2506296cfb6bcb852aa7a9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tonidero (approved):
Thank you for fixing! 🙏

PR #996, opened 2026-07-22
Add DangerousSettings support for enabling workflows

vegaro:
## What changed

Adds a `useWorkflows` flag on `DangerousSettings` (default off) so RevenueCat Workflows (multipage paywalls) can be enabled from Unity, both through the Inspector and programmatically. This is the Unity counterpart to the React Native change that added the same flag, where `dangerousSettings.useWorkflows` is reachable by the consumer.

Unity passes `DangerousSettings` to the native layer as JSON, so no wrapper or FFI signatures change — only a serialized key plus native parsing.

- `DangerousSettings.cs` gains a `UseWorkflows` field and a `DangerousSettings(autoSyncPurchases, useWorkflows)` constructor, serialized alongside `AutoSyncPurchases`. Both are `public` so the flag can be set through the programmatic `PurchasesConfiguration.Builder` path, matching RN's consumer-reachable design.
- The Android bridge builds `DangerousSettings.forWorkflows(autoSyncPurchases)` when the flag is set, otherwise the existing `new DangerousSettings(autoSyncPurchases)`.
- The iOS bridge reads the flag and builds `RCDangerousSettings` via a Swift factory (`RCPurchasesUnityDangerousSettingsFactory`) that reaches the `@_spi(Internal)` `DangerousSettings(useWorkflows:)` initializer, which is not visible to the Objective-C bridge.
- The `Purchases` MonoBehaviour gains a `[SerializeField] private useWorkflows` field under Dangerous Settings, wired into the Inspector-driven `Configure` path.

## A note on visibility

Workflows are not meant for general use, but C# has no enforced opt-in equivalent to Android's `@InternalRevenueCatAPI` or iOS's `@_spi(Internal)` — the closest built-in (`[Experimental]`) needs a compiler version Unity does not ship. The public members are therefore signaled rather than enforced: marked `[EditorBrowsable(EditorBrowsableState.Never)]` to keep them out of IntelliSense and documented as unsupported, internal-only, and slated for removal in a future release.

## Verification
- [ ] Unity Test Runner (Edit Mode) API tests
- [ ] iOS build (confirms the generated `UnityFramework-Swift.h` import resolves and the Swift factory compiles)
- [ ] Android build

The pinned PurchasesHybridCommon 18.22.2 already bundles native SDKs with workflows support (purchases-android 10.14.1 `forWorkflows`, RevenueCat iOS 5.81.1 `useWorkflows` initializer), so this compiles against current dependencies. The Unity build and Test Runner were not run in this environment, so those boxes are unchecked.

---

> [!NOTE]
> **Medium Risk**
> Changes Purchases SDK configuration at startup and enables experimental workflow paywall behavior when opted in; defaults remain off but iOS adds a Swift factory dependency that must compile in UnityFramework builds.
> 
> **Overview**
> Adds an internal **`UseWorkflows`** flag on **`DangerousSettings`** (default **off**) so Unity can turn on RevenueCat Workflows (multipage paywalls) from the Inspector or **`PurchasesConfiguration`**, aligned with other SDKs. The flag is serialized as **`UseWorkflows`** in the existing dangerous-settings JSON—no native bridge signature changes.
> 
> **C# / Inspector:** **`DangerousSettings`** gains **`UseWorkflows`**, a two-arg constructor, and JSON serialization; members are marked unsupported/internal (`[EditorBrowsable(Never)]`). **`Purchases`** wires a private **`experimentalUseWorkflows`** field (with **`FormerlySerializedAs`**) into Inspector auto-configure.
> 
> **Native:** **Android** reads **`UseWorkflows`** and uses **`DangerousSettings.forWorkflows`** when set. **iOS** parses the flag and builds settings via a new Swift **`RCPurchasesUnityDangerousSettingsFactory`** (needed for the SPI **`useWorkflows`** initializer), with **`UnityFramework-Swift.h`** imported from the Obj-C helper.
> 
> **Subtester:** **`TesterApp`** can force runtime **`Configure`** (sets **`useRuntimeSetup`** in **`Awake`**) and pass **`useWorkflows`** into **`DangerousSettings`** for internal testing; scene defaults updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1862546c3f2be31a3c01fc5fc462324bbaf72b59. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


RevenueCat/Plugins/iOS/RCPurchasesUnityDangerousSettingsFactory.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+@_spi(Internal) import RevenueCat
```
tonidero, line 2: [Unrelated to this line]
The file name seems wrong? 😅

RevenueCat/Plugins/iOS/RCPurchasesUnityDangerousSettingsFactory.swift
```diff
@@ -0,0 +1,14 @@
...
+
+/// Builds `DangerousSettings` for the Unity bridge. Lives in Swift because `useWorkflows`
+/// is an internal `@_spi(Internal)` flag that isn't reachable from the Objective-C bridge.
+@objc public class RCPurchasesUnityDangerousSettingsFactory: NSObject {
+
+    @objc public static func make(autoSyncPurchases: Bool, useWorkflows: Bool) -> DangerousSettings {
+        return useWorkflows
+            ? DangerousSettings(useWorkflows: true)
+            : DangerousSettings(autoSyncPurchases: autoSyncPurchases)
```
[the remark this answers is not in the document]
tonidero: I agree, this seems wrong

RevenueCat/Scripts/Purchases.cs
```diff
@@ -45,6 +45,11 @@ public partial class Purchases : MonoBehaviour
              "it through PurchasesConfiguration instead")]
     public bool autoSyncPurchases = true;
 
+    [Tooltip("Internal RevenueCat use only. Enables RevenueCat Workflows (multipage paywalls). " +
+             "Behavior may change without warning. Not part of the public API.\n" +
+             "NOTE: This value will be ignored if \"Use Runtime Setup\" is true.")]
+    [SerializeField] private bool useWorkflows = false;
```
tonidero: I wonder if we should name this `experimentalUseWorkflows` or something like that, so it's more visible that it is experimental from the editor... But well, not a blocker

tonidero (commented):
Some small things, but almost there. I think the only blocker is cursor's comment.

RevenueCat/Plugins/iOS/RCPurchasesUnityDangerousSettingsFactory.swift
vegaro: Ugh yeah hahaha third time I rename this file. Claude keeps insisting haha

Btw I was wondering if it would cause issues because it's swift but it works great. So we can probably add more swift files

RevenueCat/Scripts/Purchases.cs
vegaro: Oh that's a good idea. I was wondering if the comment was enough to be able to just remove it in the future. But the experimental prefix is probably super extra clear

RevenueCat/Plugins/iOS/RCPurchasesUnityDangerousSettingsFactory.swift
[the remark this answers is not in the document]
vegaro: you know I believe this is a bug in React Native as well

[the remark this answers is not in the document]
vegaro: ok, actually in RN is fine because it never receives autoSyncPurchases. But it does here... I believe I need to make an update in purchases-ios 
https://github.com/RevenueCat/purchases-ios/pull/7276/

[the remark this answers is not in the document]
tonidero: Ahhh right... I won't block the PR with this though, since this is a new property in any case and most people won't be using both properties at all

Subtester/ProjectSettings/GvhProjectSettings.xml
```diff
@@ -16,7 +16,7 @@
   <projectSetting name="Google.IOSResolver.PodToolShellExecutionSetLang" value="True" />
   <projectSetting name="Google.IOSResolver.SwiftFrameworkSupportWorkaroundEnabled" value="True" />
   <projectSetting name="Google.IOSResolver.SwiftLanguageVersion" value="5.0" />
-  <projectSetting name="Google.IOSResolver.SwiftPackageManagerEnabled" value="True" />
+  <projectSetting name="Google.IOSResolver.SwiftPackageManagerEnabled" value="False" />
```
tonidero: Hmm why this? Probably not a blocker but just curious...

tonidero (approved):
Just a question but not a blocker!

Subtester/ProjectSettings/GvhProjectSettings.xml
vegaro: I had issues when exporting to iOS. The RevenueCatUI dependency wasn't being included in the exported project. I think I had a mix of pods and SPM in the export. I am going to revert this. Maybe a clean clone of the repo fixes it instead and it will just use SPM

RevenueCat/Plugins/iOS/PurchasesUnityHelper.m
```diff
@@ -99,7 +107,9 @@ - (void)setupPurchases:(NSString *)apiKey
         NSLog(@"Error parsing dangerousSettings JSON: %@ %@", dangerousSettingsJson, error.localizedDescription);
     } else {
         BOOL autoSyncPurchases = dangerousSettingsDict[@"AutoSyncPurchases"];
-        dangerousSettings = [[RCDangerousSettings alloc] initWithAutoSyncPurchases:autoSyncPurchases];
+        BOOL useWorkflows = [dangerousSettingsDict[@"UseWorkflows"] boolValue];
+        dangerousSettings = [RCPurchasesUnityDangerousSettingsFactory makeWithAutoSyncPurchases:autoSyncPurchases
+                                                                                  useWorkflows:useWorkflows];
```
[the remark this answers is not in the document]
vegaro: I think this is no longer valid

PR #1008, opened 2026-07-24
Fix concurrent Unity callback routing

vegaro:
### Motivation

`Purchases` stores one mutable callback per operation. Two overlapping calls can overwrite each other, so the first response invokes the wrong callback and the other response is dropped.

### Description

Correlate every one-shot callback with the request IDs added by the earlier PRs in this stack.

- Register callbacks by request ID before calling the native wrapper.
- Remove and invoke only the callback matching each response.
- Keep duplicate and unknown responses from invoking callbacks twice.
- Update the existing callback fixtures to use the same request-ID response envelope as the native bridges.
- Keep the existing malformed storefront logging behavior.

No public API changes in this final layer.

### Testing

- Full EditMode suite: 145/145 passed
- Unity package export and import passed
- `./scripts/check-meta-files.sh` passed
- `./scripts/check-android-keep-annotations.sh` passed

Physical-device concurrency smoke tests were not run.

### Stack

1. #1066
2. #1067
3. #1068
4. This PR

### DIFF
--- a/IntegrationTests/Assets/CallbackTests.meta
+++ b/IntegrationTests/Assets/CallbackTests.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: e6209c54c8c6414da0098f21cb948ecc
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData:
+  assetBundleName:
+  assetBundleVariant:
--- a/IntegrationTests/Assets/CallbackTests/Editor.meta
+++ b/IntegrationTests/Assets/CallbackTests/Editor.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: 72e6daaaa28249daaf276311f1223b4a
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData:
+  assetBundleName:
+  assetBundleVariant:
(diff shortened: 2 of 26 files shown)


vegaro:
Heads-up from #1019: `_getCurrentOfferingForPlacement` carries an ordering bug that this branch preserves, and it gets worse under request-ID correlation.

`RevenueCat/Scripts/Purchases.cs:1682` checks for a missing `offering` before `ResponseHasError`. Both native wrappers send **only** an `error` key on failure — there is no `offering` key in that payload — so every failure returns early via `callback(null, null)` and the error is dropped. A network failure is indistinguishable from "no offering is configured for this placement".

Suggested change, in this branch's shape:

```csharp
private void _getCurrentOfferingForPlacement(string offeringJson)
{
    var response = JSON.Parse(offeringJson);
    if (!TryTakeCallback(response, out GetCurrentOfferingForPlacementFunc callback) || callback == null)
    {
        return;
    }

    // The error check has to come first: both native wrappers send only an "error" key on
    // failure, so testing for a missing "offering" first reports every failure as
    // "no offering configured for this placement".
    if (ResponseHasError(response))
    {
        callback(null, new Error(response["error"]));
        return;
    }
    if (response["offering"] == null)
    {
        callback(null, null);
        return;
    }
    callback(new Offering(response["offering"]), null);
}
```

#1019 makes this exact change against `main` and has a regression test for it (`GetCurrentOfferingForPlacementDeliversNativeError`), so whichever lands second will need the other's version. Not pushing it here since this branch is checked out in another workspace and currently shows conflicts with `main`.

PR #1012, opened 2026-07-24
test: cover missing Purchases functions

vegaro:
Adds `CallbackResponseTests.cs`: customer info, login, logout, restore, sync, offerings, eligibility, storefront, promotional offers, placement offerings, and Amazon LWA consent status.

---

> [!NOTE]
> **Low Risk**
> Test-only additions with no production code changes; risk is limited to test maintenance and documenting known callback edge cases.
> 
> **Overview**
> Adds **`CallbackResponseTests`** edit-mode coverage for the `Purchases` private `_*` receivers that native iOS/Android wrappers invoke via `UnitySendMessage`, using a **`PurchasesWrapperSpy`** and JSON fixtures that mirror hybrid-common wire formats (explicit nulls, error envelopes).
> 
> Tests drive success and error paths for **get products**, **offerings**, **can make payments**, **customer info**, **log in/out**, **restore/sync**, **intro eligibility**, **storefront**, **promotional offers**, **placement offerings**, and **Amazon LWA consent**, including **single-fire callbacks** on duplicate native responses and platform-specific payload shapes.
> 
> One test **documents current behavior**: **`GetCurrentOfferingForPlacement`** treats error-only payloads like “no offering” because the missing-offering guard runs before the error check.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4edc64aef4164b36f468b8f5e35e9b91bbe1ab77. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1019, opened 2026-07-28
fix: surface native errors from GetCurrentOfferingForPlacement

vegaro:
## What changed

`GetCurrentOfferingForPlacement` could never report a failure. Both native wrappers send only an `error` key when the offerings fetch fails — there is no `offering` key in that payload — but `_getCurrentOfferingForPlacement` tested for a missing `offering` first and returned early with `callback(null, null)`. The error was dropped, so a network failure looked exactly like "no offering is configured for this placement".

Moving the error check ahead of the missing-offering guard fixes it.

## Base branch

Stacked on `cesar/unity-callback-response-tests` (#1012), which adds the test file this builds on.

---

> [!NOTE]
> **Medium Risk**
> Changes callback semantics for placement offerings on the Unity bridge; apps that relied on `(null, null)` for failures may need to handle `Error`, but the fix aligns behavior with other offerings APIs and is covered by tests.
> 
> **Overview**
> **`GetCurrentOfferingForPlacement`** now forwards native failures to the app callback instead of treating them like an empty placement.
> 
> In **`_getCurrentOfferingForPlacement`**, **`ResponseHasError`** runs **before** the missing-`offering` guard. On failure, iOS/Android only send an **`error`** key (no **`offering`**), so the old order always invoked **`callback(null, null)`** and hid network and other errors.
> 
> The edit-mode test **`GetCurrentOfferingForPlacementDeliversNativeError`** now expects a populated **`Error`** (e.g. code 10, **`NETWORK_ERROR`**) instead of documenting dropped errors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c72bdc79fd70cc7bab5140c4059eeafaf825c467. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/IntegrationTests/Assets/Tests/EditMode/CallbackResponseTests.cs
+++ b/IntegrationTests/Assets/Tests/EditMode/CallbackResponseTests.cs
@@ -617,7 +617,7 @@ namespace RevenueCat.Tests
         }
 
         [Test]
-        public void GetCurrentOfferingForPlacementDropsNativeError()
+        public void GetCurrentOfferingForPlacementDeliversNativeError()
         {
             Purchases.Offering receivedOffering = CreateOffering();
             Purchases.Error receivedError = null;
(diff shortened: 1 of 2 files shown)


tonidero (approved):
Thank you!!

PR #1023, opened 2026-07-29
Resolve the test app native dependencies with EDM4U

ajpallares:
## Summary

The test app resolved its native dependencies two bespoke ways: a hand-written Swift Package hook for iOS, and a checked-in snapshot of resolver output for Android. The snapshot had drifted to PurchasesHybridCommon 18.1.0 while the SDK declares 18.25.0, so Android was silently testing a much older native SDK than iOS.

- Both platforms now resolve from the SDK's own `*Dependencies.xml`, the way a customer project does, which is also what this suite is meant to exercise.
- The Android build fails if the resolved version stops matching the SDK's, so this can't drift unnoticed again.

### Notes
- I could not verify locally: this machine has no Android module, and batch-mode player builds fail licence activation. CI is the first real run.
- Unverified: whether EDM4U resolves during a batch-mode build at all. If it doesn't, the new check fails loudly rather than silently.

Depends on #838.


ajpallares:
Closing: the problem this PR was opened for is fixed more simply in #836.

The Android drift didn't need EDM4U. The Gradle block is ordinary Maven
coordinates that Gradle resolves at build time, and the real cause was that
the test app's `mainTemplate.gradle` was never registered with the version
bump lane, so it stayed on 18.1.0 while every registered file moved to
18.26.0. #836 now registers it alongside `Subtester`'s, and brings the pin
up to the declared version first, since `replace_in` gsubs the current
version string and would otherwise skip an already-stale file silently.

This PR emptied the dependency block, which manufactured a dependency on
the resolver that never needed to exist. The only remaining rationale was
replacing the hand-rolled iOS Swift Package hook, which works today and
whose batch-mode resolution was never confirmed on CI. Not worth the
vendored resolver and slower Android builds for that alone.

PR #1066, opened 2026-09-01
Add callback request registry

vegaro:
### Motivation

This is the first part of #1008 which fixes the fact that `Purchases` stores one mutable callback per operation. Two overlapping calls can overwrite each other, so the first response invokes the wrong callback and the other response is dropped.

### Stack

1. This PR
2. #1067
3. #1068
4. #1008

---

> [!NOTE]
> **Low Risk**
> New internal helper and tests only; no wiring into purchase flows yet.
> 
> **Overview**
> Introduces an internal **`CallbackRegistry`** so async operations can stash callbacks under unique request IDs instead of sharing a single mutable slot (foundation for fixing overlapping `Purchases` calls in later stack PRs).
> 
> **`Register<T>`** returns a GUID string and stores the callback (including null) with its type; **`TryTake<T>`** removes and returns the callback only when the ID exists and the type matches—wrong types leave the entry intact. **`Clear`** drops all pending entries. Access is guarded with a lock.
> 
> Adds **EditMode NUnit tests** for out-of-order take, single-use IDs, null callbacks, type mismatch, and clear behavior, plus Unity `.meta` assets.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ee34e882489589c9a903d813a3e9a33df645a06e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


facumenzella (approved):
nice work 🙌

heads up, the three red jobs (build-subtester-android, build-subtester-ios, build-integration-tests-ios) all failed on `unity/prepare-env`,

